PR #15, opened 2026-07-05
docs: add `npx skills add` install method to README

salmon-21:
Add `npx skills add` as the primary installation method for the shipped `SKILL.md`, so users of any supported coding agent can install the skill in one command.

- Add `npx skills add ubermenchh/anki-cli` as the recommended install method (via the [skills](https://github.com/vercel-labs/skills) CLI, which auto-detects installed agents)
- Replace the Cursor-only manual instructions with an agent-neutral skill-directory table (Claude Code, Codex, Cursor, OpenCode, project-level)
- Link `SKILL.md` and the [agentskills.io](https://agentskills.io) spec

No changes to `SKILL.md` or CLI behavior — documentation only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,18 +217,31 @@ Exit codes:
 
 ## AI Agent Integration
 
-This repo ships a `SKILL.md` at the project root that AI agents (Cursor, Windsurf, Cline, etc.) can consume to operate `anki-cli` without human guidance. It covers the full command reference, JSON output structure, search query syntax, error codes, and common workflows.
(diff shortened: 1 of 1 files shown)


ubermenchh (approved):
Thanks for this. Verified the main claim: `npx skills add ubermenchh/anki-cli` clones the repo, discovers the root `SKILL.md` as one skill, and installs it.

One non-blocking note on the manual table: OpenAI's current Codex docs list the user-level location as `~/.agents/skills/` and no longer mention `~/.codex/skills/` (that's the legacy path). Worth switching the Codex row to `~/.agents/skills/anki-cli/SKILL.md`, either here or in a follow-up. Since Codex, Cursor, and OpenCode all read `~/.agents/skills/`, the table could also collapse to that one universal row plus Claude Code.

Optionally add `-g` to the recommended command so it installs globally without prompting for project vs. global.

Issue #39, opened 2026-09-12
Tracking: codebase review findings (Sep 2026)


ubermenchh:
Tracking issue for the codebase review done on 2026-09-12 (v0.1.5). Each item is its own issue with `file:line` references, a reproduction, and a suggested fix. Suggested order of attack is top to bottom.

### 1. Data safety (direct SQLite backend) — fix first
- [x] #16 Deleting a filtered deck hard-deletes visiting cards and notes
- [x] #17 `remove_notetype_field` shifts every note's field values
- [x] #18 `col.mod` never updated; `col.scm` not bumped on schema changes
- [x] #19 `queue=3` (day-learn) treated as epoch seconds
- [ ] #20 `answer_card` / `reset_cards` ignore filtered-deck `odid/odue`
- [ ] #21 Day index uses UTC midnight instead of rollover hour
- [ ] #22 FSRS fidelity gaps (step reset, FSRS-5 params, fuzz, `reqs`, ORDER BY)
- [ ] #23 `add_note`: `allow_duplicate` not enforced; `csum` hashes raw HTML
- [ ] #24 No schema-version guard for legacy collections

### 2. Agent contract (the "for agents" promise)
- [ ] #25 `preprocess_argv` mangles option values containing `=`
- [ ] #26 `--yes` / `--format` only accepted before the subcommand
- [ ] #27 Central exception → JSON envelope mapper
- [ ] #28 Docs vs CLI drift (`cards` is a TUI, `note:bulk` schema, exit codes, flag names)

### 3. Tests (do alongside 1 and 2)
- [ ] #37 `conftest.py` with faithful Anki DDL; dedupe 16 store builders + 11 CLI helpers
- [ ] #38 Write guard, FSRS mapping, rollback never executed; checksum oracle mirrors impl

### 4. Architecture
- [ ] #29 `@anki_command` decorator; dedupe card rendering
- [ ] #30 Backend Protocol parity (shapes, `_store` methods, search semantics)
- [ ] #31 Split `anki_direct.py` into `db/` modules; fix import cycle
- [ ] #32 Lazy detection + AnkiConnect HTTP hygiene
- [ ] #33 Dead code, duplicated defaults, repo hygiene

### 5. TUI
- [ ] #34 Literal `\n`, phantom hint keys, undo stack/revlog issues
- [ ] #35 Synchronous IO on the event loop, N+1 fetches, polling
- [ ] #36 REPL re-implements dispatch and the review loop

[...]

Issue #23, opened 2026-09-12
add_note: allow_duplicate=False not enforced; csum hashes raw HTML unlike Anki


ubermenchh:
## Problem

`add_note` in `anki_cli/db/anki_direct.py`:

1. **`allow_duplicate=False` is not enforced.** `:1553-1566` looks up duplicates, writes a warning to stderr, then `:1575` inserts anyway. The AnkiConnect backend raises on duplicates, so the same command behaves differently per backend.
2. **`csum` is computed over raw HTML.** `:2250-2252` hashes the raw first field. Anki hashes `strip_html_preserving_media_filenames(field)`. So `<b>cat</b>` gets a different checksum than Anki would give it → CLI-created notes are never detected as duplicates of Anki-created ones (and vice versa), and Check Database rewrites the CLI's value.
3. The duplicate lookup at `:1556` is not scoped by `mid` (notetype), unlike Anki.
4. `csum` is computed twice from the same input (`:1553`, `:1588`).
5. `sfld` should use the same HTML-stripped text.

## Reproduce

```sh
anki note:add --deck Default --notetype Basic --Front hola --Back hello
anki note:add --deck Default --notetype Basic --Front hola --Back hello   # succeeds again on direct; fails on AnkiConnect
```

## Suggested fix

```python
def _strip_html_preserving_media(s: str) -> str:
    # keep [sound:x] and <img src="x"> filenames, drop other tags/entities
    ...

csum = sha1(_strip_html_preserving_media(first_field).encode()).hexdigest()[:8]
dup_ids = conn.execute("SELECT id FROM notes WHERE csum = ? AND mid = ?", (int(csum, 16), mid)).fetchall()
if dup_ids and not allow_duplicate:
    raise ValueError("duplicate note")
```

Test oracle: use **literal `csum` values captured from a real Anki collection** for `"Q"`, `"<b>Q</b>"`, and `<img src="x.png">` — not a re-implementation of the hash (see `tests/integration/test_anki_direct_note_mutators.py:13-15`, which currently mirrors the implementation).

umangk-2410:
Picking up the remaining half of this (allow_duplicate enforcement, mid-scoped dup lookup, single csum computation). The csum HTML-strip is in #50; I'll leave sfld stripping until that lands so we don't both touch `_field_checksum`.

Issue #22, opened 2026-09-12
FSRS fidelity gaps: step reset, FSRS-5 params, fuzz mismatch, reqs ignored, ORDER BY unit mix


ubermenchh:
Several smaller fidelity gaps between the direct-backend FSRS scheduler and Anki's. None corrupts the collection on its own, but each produces intervals or queue states that differ from what Anki would write for the same rating.

### 1. Learning step is always reset to 0
`anki_cli/db/anki_direct.py:2554-2555` — `step = 0 if left_raw > 0 else None` discards `left % 1000` (remaining steps). `_map_fsrs_result_to_anki` (`:2596-2599`) then re-encodes from `next_card.step`.

Repro: learn steps `[1, 10, 60]` min, card `type=1, queue=1, left=1001` (one step left), rate Good. Anki graduates the card to review. CLI: FSRS step 0→1, `due = +10m`, `left = 2002` — two extra steps.

Fix: `remaining = left_raw % 1000; step = max(0, step_count - remaining)`, passing the deck's learn/relearn step counts into `_card_row_to_fsrs`.

### 2. FSRS-5 (19-param) weights fed to the FSRS-6 scheduler
`:2470-2473` accepts `len(values) >= 19`. Collections last optimized on Anki 24.06–24.10 have `fsrs_params_5` populated and `fsrs_params_6` empty; `fsrs` 6.x reads `parameters[20]` for decay → `IndexError`/`ValueError` on every `review:answer` in that deck.

Fix: accept exactly 21; for 19 or 17 either upgrade with FSRS's documented defaults for the missing weights or fall back to `DEFAULT_PARAMETERS` and warn.

### 3. Fuzz mismatch between preview and answer
`:2460` — py-fsrs fuzzing is on by default. `preview_ratings` and `answer_card` build separate schedulers, so the interval shown can differ from the one written.

Fix: `enable_fuzzing=False` for preview, or seed identically.

### 4. Revlog row fidelity
`:2070`, `:2051-2056` — `revlog.time` is always 0; a relearning (`type=3`) card is logged as `type=0` (learn) instead of `2` (relearn).

### 5. Non-Anki key in `cards.data`
`:1993` — `data["lrt"]` is not an Anki `CardData` key; Anki drops it on its next write, so it's an unreliable cache. Remove or move to a side table.

### 6. `_get_next_due_for_deck` ORDER BY mixes units
[...]

Issue #18, opened 2026-09-12
col.mod never updated; col.scm not bumped on notetype schema changes -> sync sees 'No changes'


ubermenchh:
## Problem

Every write path in `anki_cli/db/anki_direct.py` sets `usn = -1` and per-row `mod`, but **`col.mod` is never updated** (`rg "UPDATE col" anki_cli/db/anki_direct.py` → 0 hits), and `col.scm` is never bumped on notetype schema changes.

Consequences:

1. Anki's normal-sync handshake compares local `col.mod` with the server. If equal → `NoChanges`; the pending `usn = -1` rows are not even scanned. So reviews/edits made via the CLI silently do not upload until some unrelated desktop-side edit nudges `col.mod`. Reviews of the same cards on another device win silently in the meantime.
2. `add_notetype_field`, `remove_notetype_field`, `add_notetype_template` (`:407-413`, `:447-454`, `:501-513`) only touch `fields`/`templates` rows. Anki syncs notetypes as whole objects keyed off `notetypes.usn = -1`, and treats field/template add/remove as a **schema change** requiring a `col.scm` bump (forced one-way sync). Today other devices never receive the new field, and later CLI-edited notes carry N+1 values against an N-field notetype → "field count mismatch" in Check Database.

## Reproduce

```sh
# answer 20 cards
anki review:answer --id ... --rating good  # x20
# open Anki, click Sync immediately
```

Anki reports "No changes".

## Suggested fix

- In `_connect_write` (`:61-78`), before `COMMIT`:
  ```python
  conn.execute("UPDATE col SET mod = ?", (int(time.time() * 1000),))
  ```
- After any `fields`/`templates` mutation:
  ```python
  conn.execute("UPDATE notetypes SET mtime_secs = ?, usn = -1 WHERE id = ?", ...)
  conn.execute("UPDATE col SET scm = ?, mod = ?", (now_ms, now_ms))
  ```
  and print a warning that the next sync will be a full upload.

Add assertions to the shared test fixture (see #37) that `col.mod` increases after every mutator.

PR #40, opened 2026-09-12
fix(direct): stop data loss on filtered-deck delete and notetype field removal

### LINKED ISSUE #16: Deleting a filtered deck hard-deletes visiting cards and notes
## Problem

`delete_deck` (`anki_cli/db/anki_direct.py:1356-1432`) never reads the deck's `kind` or the cards' `odid`. Deleting a **filtered deck** therefore hard-deletes every card that is currently *visiting* it, deletes the notes whose cards were all in scope, and writes graves — so the deletion propagates to every synced device.

Anki's own behavior for a filtered deck is to return the cards home (`did = odid`, `due = odue`, `odid = 0`, `odue = 0`) and delete only the deck row.

It also does not refuse to delete deck id `1` (Default), which Anki forbids.

## Reproduce

Collection with deck `{"id": 555, kind: filtered}` and 30 cards `did=555, odid=1, odue=19700`.

```sh
anki --yes deck:delete --deck "Cram"
```

Result: 30 cards + their notes deleted and graved. Expected: 30 cards moved back to deck 1, 0 notes deleted.

## Suggested fix

```python
kind_name, _ = betterproto.which_one_of(deck_kind, "kind")
if kind_name == "filtered":
    conn.execute(
        "UPDATE cards SET did = odid, due = odue, odid = 0, odue = 0, mod = ?, usn = -1 WHERE did = ?",
        (now, did),
    )
    # delete only the deck row + grave type 2
else:
    # existing path
if did == 1:
    raise ValueError("Cannot delete the Default deck")
```

Add an integration test with a filtered deck fixture asserting the cards survive with `did = odid`.

### LINKED ISSUE #17: remove_notetype_field shifts every note's field values (never rewrites notes.flds)
## Problem

`remove_notetype_field` (`anki_cli/db/anki_direct.py:447-454`) deletes the `fields` row and decrements higher ords, but **never rewrites `notes.flds`**. Every existing note of that notetype now has its field values shifted by one relative to the schema. `sfld`, `csum`, and `reqs[].field_ords` are also left stale, and neither `notetypes.usn` nor `col.scm` is bumped (see #18).

## Reproduce

Notetype fields `[Front(0), Extra(1), Back(2)]`, note `flds = "a\x1fb\x1fc"`.

```sh
anki notetype:field:remove --notetype MyType --field Extra
```

Schema becomes `[Front(0), Back(1)]`. `get_note_fields` now returns `{"Front": "a", "Back": "b"}` and Anki renders `{{Back}}` as `b`. Expected `c`.

## Suggested fix

In the same `_connect_write` transaction:

1. `SELECT id, flds FROM notes WHERE mid = ?`
2. For each: split on `\x1f`, `pop(removed_ord)`, rejoin.
3. Recompute `sfld` and `csum` (HTML-stripped, see #23).
4. `UPDATE notes SET flds=?, sfld=?, csum=?, mod=?, usn=-1 WHERE id=?`
5. Drop `removed_ord` from every `reqs[].field_ords` in the notetype config and renumber higher ords.
6. `UPDATE notetypes SET mtime_secs=?, usn=-1`, and bump `col.scm` (schema change → forces full sync).

Add an integration test that inserts a 3-field note, removes the middle field, and asserts `flds == "a\x1fc"`.

ubermenchh:
Fixes the two BLOCKING data-safety bugs from the review tracker (#39).

Closes #16
Closes #17

## #16 `delete_deck` hard-deleted cards visiting a filtered deck

`anki_cli/db/anki_direct.py:1350`

- Reads each deck's `kind` and splits filtered vs normal.
- **Filtered**: new `_return_cards_from_filtered_decks` sets `did = odid, due = odue, odid = 0, odue = 0`, recomputes `queue` from `type` (suspended/buried cards keep their negative queue), bumps `mod`/`usn`. Only a type-2 grave is written. Mirrors Anki's `remove_from_filtered_deck_restoring_queue`.
- Cards in a filtered deck with `odid = 0` (malformed) are parked in Default instead of dangling.
- **Normal**: scope is `did IN (...) OR odid IN (...)`, so cards owned by the deck but on loan to a filtered deck are deleted too (Anki's `delete_all_cards_in_normal_deck`).
- Refuses to delete deck id 1.
- Result gains `returned_cards`.

## #17 `remove_notetype_field` never rewrote `notes.flds`

`anki_cli/db/anki_direct.py:417`

- New `_remove_field_from_notes` rewrites every note of the notetype in the same transaction: pads short rows, pops the removed slot, recomputes `sfld`/`csum`, sets `mod`/`usn = -1`.
- `sort_field_idx` shifts correctly (decrement if after; cap to last field if it was the removed one, like Anki's `reposition_sort_idx`).
- `reqs[].field_ords` renumbered; empty requirements become `KIND_NONE`.
- `notetypes` row always marked modified.
- Result gains `updated_notes`.

## Tests

8 new integration tests (599 → 607 passing). Fixtures for both files gained the columns needed (`cards.odid/odue/type/queue/due/mod/usn`, a `notes` table for the notetype fixture). Each commit passes on its own.

## Deliberately not in this PR

- `col.mod` / `col.scm` bump → #18 (belongs in `_connect_write` so every mutator gets it).
- `csum` HTML stripping → #23.
- Noticed while here: `create_notetype` writes `reqs` as `KIND_ALL` over every field, which would suppress card generation unless *all* fields are filled. Adding to #22.


umangk-2410:
Addressed in 1db2b54.

**Important**
- `deck:delete` now maps `ValueError` → `INVALID_INPUT`, exit 2 (`cli/commands/deck.py`); CLI test added.
- `notetype:field:remove` requires `--yes` (`cli/commands/notetype.py`); CLI test added; README/SKILL.md updated and the command added to the destructive list.
- Reqs/csum test: added a `KIND_ALL [0, 2, 3]` requirement so ords past the removed one must shift, and asserted `csum == checksum(first field)` while `sort_idx != 0`.
- Filtered-deck test: added a new card with `odue == 0` (due must be left alone) and a buried card (`queue -3` must survive); `returned_cards` is now 6.

Re-ran the four named mutations against the two integration files: each now fails exactly one test (`ord_ - 1 …` → `ord_`; csum from sort field; `restored_due` → `"odue"`; `queue < 0 THEN queue` → `-1`).

**Nits**
- Dropped the redundant `elif sort_idx == removed_ord` branch; comment moved above the clamp.
- `delete_deck` fails closed with `ValueError` on a deck kind that is neither `normal` nor `filtered`; test with an empty kind blob asserts nothing is touched.
- Reverted the `executescript(` reflow.

**Suggestions taken here**
- Normal-deck deletion moved into `_delete_cards_owned_by_decks`, staging ids in `TEMP` tables: graves come from the same set that was deleted, no `IN (?, …)` list, and the `NOT IN (SELECT … WHERE NOT scope)` full scan is gone.
- `remove_notetype_field` matches field names with `casefold()` (test added).

**Filed as follow-ups**
- #42 `add_notetype_field` never appends a slot to `notes.flds`
- #43 queue-from-type CASE ×4, three of them wrong for type 3
- #44 chunk the notes rewrite; recompute `reqs` conservatively

Haven't run against a real `collection.anki2` for the `(ntid, ord) WITHOUT ROWID` update order or `EXPLAIN QUERY PLAN` on the scope query yet — will do that before merging.

umangk-2410 (approved):
Re-reviewed `1db2b54` + `3fee80d` against the earlier findings.

- All four Important items resolved; ran the mutant matrix rather than trusting the claim — M1–M4 each fail exactly one test now.
- New `_delete_cards_owned_by_decks` checks out: orphan-note predicate is equivalent to "every card of the note is deleted", graves come from the same snapshot as the deletes, and the real Anki schema has no FKs so the notes-before-cards order is free.
- One new survivor found (`sort_idx >` → `>=`, the "sort field itself removed, not last" case) — pinned in `3fee80d` with a test matching Anki's `reposition_sort_idx`; N1/N3–N6 all killed.
- Verified on a real schema-18 `collection.anki2`: the `WITHOUT ROWID` ord shift is collision-free, and the `odid != 0` guard added in `3fee80d` moves the scope query from `SCAN cards` to `MULTI-INDEX OR` over `ix_cards_sched` + `idx_cards_odid`.

Follow-ups already filed: #42, #43, #44. The unescaped `LIKE 'X::%'` subtree match (`_`/`%` in deck names) is pre-existing across four sites and worth its own issue, not a blocker here.

CI green, 613 tests. Approving.

Issue #37, opened 2026-09-12
tests: add conftest.py with faithful Anki DDL; remove 16 store builders and 11 CLI helper copies


ubermenchh:
## Problem

There is **no `conftest.py` anywhere** in `tests/`. As a result:

- **16 private `_make_store()` builders**, one per `tests/integration/*.py`, each with a *different partial schema*. `col` often has only `crt` (real: 13 columns); `decks` in `test_anki_direct_read_apis.py:45` lacks `mtime_secs`/`usn`; `test_anki_direct_note_mutators.py:34-49` inserts `notetypes.config = b""` and `templates` with no `name`/`config`.
- **11 copies** of the CLI helper quartet (`_base_obj` / `_success_payload` / `_error_payload` / `_patch_session`) across `tests/cli/*.py`.
- Because each file's schema omits different columns, a regression in sync bookkeeping (`mod`, `usn`, `mtime_secs`, `col.mod`) is invisible in whichever file happens to omit that column.
- 17× `{k: row[k] for k in row.keys()}` (ruff SIM118). Note the naive fix `for k in row` is **wrong** for `sqlite3.Row` (iterates values); use `dict(row)`.
- `integration` / `unit` / `slow` markers are registered in `pyproject.toml:80-85` but never applied.
- `tests/unit/test_entrypoint_and_colors.py:21-38` asserts color constants are hex strings — tests nothing a bug would break.
- `tests/backends/test_ankiconnect.py:103-106` — first `_backend_with_handler` assignment is dead, immediately overwritten.

## Suggested fix

`tests/conftest.py`:

```python
@pytest.fixture
def collection(tmp_path) -> Collection:
    """Real Anki schema-18 DDL for all 13 tables, with a Default deck, Basic notetype, and default deck_config."""
    ...

# builders that default every column:
collection.insert_deck(name=..., **overrides)
collection.insert_note(fields=[...], **overrides)
collection.insert_card(nid=..., **overrides)
collection.row(table, id) -> dict   # dict(row)
collection.col() -> dict
```

Copy the DDL from Anki's `rslib/src/storage/schema18_upgrade.sql` (or dump `.schema` from a fresh modern collection) so the fixture is faithful.

`tests/cli/conftest.py`: move the four CLI helpers there.

Then:
- Delete the 16 private builders and 11 helper copies.
[...]

Issue #32, opened 2026-09-12
Lazy backend detection + AnkiConnect HTTP hygiene (Connection: close, timeouts, N+1, param fallbacks)


ubermenchh:
### Detection runs eagerly and twice
`cli/app.py:132-146` calls `detect_backend` on **every** invocation, including `config get` and `shell`: HTTP probe (0.35 s) + `pgrep` process scan + SQLite lock probe. Then `factory.py:33-40` constructs `AnkiConnectBackend(verify_version=True)`, which posts `version` again with a fresh client.

- If Anki is open without the AnkiConnect add-on, `anki config get` exits 7.
- `detect.py:95-99` returns `standalone` as the fallback, but `factory.py:52-55` always raises `BackendNotImplementedError` for it, so the "nothing found" path fails late with a misleading "detected but not implemented".
- `detect.py:59, 83` checks `_anki_process_running()` (shells out to `pgrep` on macOS) *before* the cheaper, more accurate `_sqlite_write_locked()`.
- `detect.py:114-115` — `--col` pointing at a missing file silently returns `None` and falls through to auto-discovery of a *different* collection.
- `detect.py:132` picks `candidates[0]` alphabetically when multiple profiles exist; `collection.anki_profile` config exists but is never read.

### HTTP client hygiene (`backends/ankiconnect.py`)
- `:54, :98` — an `httpx.Client` is created for reuse, then every request sends `Connection: close`, defeating pooling. With the N+1 patterns (`get_notetypes` = 1 + 2·models, `get_tag_counts` = 1 + tags, `get_due_counts` = 3, `set_card_flag` = 1/card), `notetype list` on 30 models is 61 TCP handshakes.
- `:743-749` — one 2.0 s timeout covers connect *and* read; a large `findCards` surfaces as "Timeout contacting AnkiConnect". No retry on `RemoteProtocolError` (likely the real reason `Connection: close` was added in e7fe2a5).
- `:179-225` — `rename_deck` tries 3 param spellings, then emulates via create/changeDeck/delete (non-atomic: a mid-loop failure leaves duplicated decks).
- `:428-435, :439-446, :609-613, :685-699` — `edit_notetype_template`, `set_notetype_css`, `answer_card`, `unbury_cards` each try two param shapes catching `AnkiConnectAPIError`, which also masks real errors ("model not found" triggers the second attempt). `check_version` already pins ≥ 6, so one documented shape suffices.

## Suggested fix

[...]

Issue #21, opened 2026-09-12
Day index uses UTC midnight instead of Anki's local rollover hour


ubermenchh:
## Problem

Anki computes `days_elapsed` from `col.crt` adjusted to the **local timezone plus the `rollover` hour** (default 4 am, stored in `col.config`). `anki_cli/db/anki_direct.py` floors UTC epochs instead, in five places:

- `:2106-2114` (`_today_days` or equivalent)
- `:2544-2545`
- `:2581-2582`
- `:2170`
- `:2774`

## Reproduce

`crt = 1699952400` (2023-11-14 04:00 EST), now `1700028000` (2023-11-15 01:00 EST).

- CLI: `19676 - 19675 = 1` → "day 1".
- Anki: 21 h since rollover → "day 0".

Cards due "tomorrow" appear due now; `answer_card` on a 1-day interval writes `due = crt_day + 2`, which Anki shows as the day after tomorrow.

## Suggested fix

- One shared helper `_day_index(epoch_s: int, crt_s: int, rollover_h: int, tz) -> int`, used at all five sites.
- Read `rollover` from `col.config` (protobuf `ConfigKey` / legacy JSON key `rollover`), default 4.
- Compute via `datetime.fromtimestamp(now, local_tz) - timedelta(hours=rollover)`, then take the date difference.

Table-driven unit test with a fixed clock covering: before rollover, after rollover, DST boundary.

Issue #7, opened 2026-02-25
Hotkey to show the keys


ilya-grigoriev:
How about entering a hotkey `?` to show the keys as in `htop`?

umangk-2410:
Planned as part of #84: bind `?` to Textual's built-in `action_show_help_panel` and replace the hand-rolled hint bars with `Footer`. Tracked in #88.

PR #41, opened 2026-09-12
refactor: sweep dead code, duplicated defaults, and unused config fields

### LINKED ISSUE #33: Dead code, duplicated defaults, unused config fields, repo hygiene
A sweep of dead or duplicated definitions. Each is small; grouped here so they can be knocked out in one PR.

### Config fields that are never read
`models/config.py:8, 20, 23-31` — `collection.anki_profile`, `display.day_boundary_hour`, `backup.*`, `review.show_timer`, `review.max_answer_seconds` are definable via `config set` and serialized (`config_runtime.py:350`) but no code reads them. Either wire them (`anki_profile` → `_resolve_direct_collection`, `day_boundary_hour` → the rollover fix) or delete.

### Triplicated defaults
- `"http://localhost:8765"` in `backends/detect.py:12`, `backends/factory.py:27`, `models/config.py:13`.
- `~/.local/share/anki-cli/collection.db` in `detect.py:145` and one other place.
- `JSONValue` defined in both `backends/protocol.py:7-8` and `models/output.py:8-9`.

### Hand-rolled TOML writer
`config_runtime.py:347-361` — `_toml_scalar` throws on `list` values, but `_coerce_raw_value` (`:317`) accepts JSON lists. Use `tomli_w` or reject lists at set time.

### Unused symbols
- `tui/repl.py:90-91` — `_TAG_RE`, `_BR_RE` (unused; `_BR_RE` is also wrong: `r"(?i)<br\\s*/?>"` matches `<br\s>` not `<br/>`).
- `tui/repl.py:305` — `_due_counts_inline`.
- `tui/review_app.py:87` — `_queue_name`.
- `tui/browse_app.py:59` `_COLUMNS`, `:94` `_format_card_row`, `:793` `_count_for_filter`.
- `_relative_eta` duplicated in `review_app.py:100` and `browse_app.py:170`.
- `cli/app.py:61-62` — `--quiet` / `--verbose` stored but never read.
- `cli/commands/general.py:52-69` — `init` returns `"implemented": False`; `:47` `status` returns `"message": "foundation in progress"`.
- `standalone` backend selectable at `app.py:57` but always raises (`factory.py:52-55`).
- `backends/detect.py:274-288` — `_sqlite_write_locked` is fine; but `_anki_process_running` is checked first (see #32).

### Repo hygiene
- `.DS_Store` untracked at repo root; add to `.gitignore`.
- `__pycache__` directories present in the working tree (already ignored, just noting).
- CI runs `ruff check anki_cli` only; `tests/` has 17× SIM118 + 1 E501. Run `ruff check .` in CI (see #37 for the `sqlite3.Row` caveat).
- Enable `ruff --select W` (trailing whitespace / EOF newline: ~19 hits in cli+tui).

LouisDeconinck:
Closes #33

## Summary

Implements the cleanup checklist:

- **Unread config fields** — `collection.anki_profile` is now wired into `detect_backend()` (prefers the candidate whose parent dir matches the configured profile); `display.day_boundary_hour`, `backup.*`, `review.show_timer`, `review.max_answer_seconds` deleted along with their models. (`day_boundary_hour` deleted rather than wired — wiring it is #21's local-rollover math; that issue should re-add the field.)
- **Triplicated defaults** — single `DEFAULT_ANKICONNECT_URL` shared by `detect.py`/`factory.py`; duplicated collection-path and `JSONValue`/`JSONPrimitive` definitions consolidated.
- **TOML writer** — dead list/dict JSON branches removed so non-scalars are rejected at set time; serializer iterates the model dump instead of a hardcoded section list.
- **Unused symbols** — dead helpers/imports/flags removed; `--quiet`/`--verbose`, the `init` no-op stub, and `status`'s stale `message` field dropped; `standalone` backend removed end-to-end (auto-detect now raises `DetectionError(exit_code=3)` per #32's direction).
- **Repo hygiene** — `.DS_Store` in `.gitignore`; CI runs `ruff check .`; `"W"` added to ruff select (156 violations auto-fixed); SIM118 fixed via `dict(row)` per #37's `sqlite3.Row` caveat.

Left `_anki_process_running`/`_sqlite_write_locked` ordering alone — the issue marks it "(see #32)" and #32 owns that reorder.

## Verification

- `ruff check .` clean; `ty check anki_cli` clean
- `pytest`: 594 passed; `pytest -m tui`: 51 passed
- Smoke: `--backend standalone` rejected, `init` unregistered, `config set` writes valid TOML

Note: `uv.lock` upstream is stale (locks 0.1.4 vs pyproject 0.1.5) — reverted the regeneration to keep the diff clean; any local `uv run` without `--frozen` will rewrite it.


umangk-2410 (dismissed):

ubermenchh (changes requested):
Thanks for taking #33 on — this is a careful sweep, and the parts that are pure deletion are done right (see the praise at the bottom). Two things block it, though, and both were reproduced locally on this branch.

**Local gates:** `ruff check .` clean, `ty` clean, 594 + 51 tui tests pass.

## 🔴 Blocking

### 1. `config:set` persists the dead standalone `collection.path` default → total lockout; the new `anki_profile` feature is unreachable

`anki_cli/models/config.py:11` · `anki_cli/config_runtime.py:228-229, 331-341` · `anki_cli/backends/detect.py:110-112, 90-93`

`CollectionConfig.path` still defaults to `~/.local/share/anki-cli/collection.db` (the deleted standalone location). `_serialize_config_toml` dumps the full model, so *any* `config:set` writes that default into the file. On the next run `_resolve_collection_override` treats the file's `collection.path` as an explicit `--col`; `detect.py:110` finds it doesn't exist and returns `None` before `anki_profile` is ever consulted; auto mode then raises exit 3.

Reproduced (macOS, profile `Work/collection.anki2` exists, Anki closed):

```
$ anki --backend direct status
  → ok, backend=direct, .../Anki2/Work/collection.anki2
$ anki --backend direct config:set --key collection.anki_profile --value Work
  → ok
# config.toml now contains:  path = "~/.local/share/anki-cli/collection.db"
$ anki --backend direct status
  → BACKEND_UNAVAILABLE "no Anki collection DB was found", exit 3
```

Every subsequent command fails, including `config:set` to undo it. The root cause predates this PR (on `main` the same sequence degraded to `standalone` in auto mode), but removing the fallback makes it fatal, and it means the feature this PR wires up can't be enabled through the CLI.

Suggested fix:

```python
# models/config.py
class CollectionConfig(BaseModel):
    path: str | None = None
    anki_profile: str | None = None   # see Important #3

# config_runtime.py:_serialize_config_toml — skip None leaves
if value is None:
    continue
```

Update `tests/unit/test_models.py:17,32` (they pin the old default) and add a round-trip test: `config:set anki_profile` → reload → `status` still resolves the collection.

### 2. Backend-less commands now exit 3 when no Anki is present

`anki_cli/cli/app.py:126-141` · `anki_cli/backends/detect.py:90-93`

`detect_backend` runs in the group callback for every subcommand. With the infallible `standalone` result gone, auto mode raises before dispatch.

Reproduced with an empty `$HOME` and no AnkiConnect: `anki version` → exit 3, `anki status` → exit 3, `anki config:path` → exit 3, bare `anki` (REPL) → exit 3. All exit 0 on `main`. The README's remote-AnkiConnect flow (`config:set backend.ankiconnect_url …` on a box with no local Anki) is now impossible, and `status` — which SKILL.md advertises as the health check — can no longer report "no backend found".

Minimal fix until #32's lazy detection lands:

```python
_BACKENDLESS = {"version", "status", "config", "config:path", "config:set", None}
if ctx.invoked_subcommand in _BACKENDLESS:
    ctx.obj.update({"collection_path": runtime.collection_override,
                    "backend": "none", "backend_reason": "not required"})
else:
    detection = detect_backend(...)
```

Plus the test in Important #7 (it fails today).

## 🟠 Important

**3. Unmatched `anki_profile` silently falls back to a *different* profile; default `"User 1"` means it's always "set"** — `detect.py:132-137`, `models/config.py:12`. Profiles `Personal/`, `User 1/`; config `anki_profile = "user 1"` (case slip; comparison is case-sensitive even on macOS) → loop matches nothing → `candidates[0]` → `note:add` writes into **Personal**. Anyone without a `User 1` profile hits this path on every run. Fix: default `None`; if set and unmatched, `raise DetectionError(f"Anki profile '{p}' not found. Available: …", exit_code=3)`. `test_detect_helpers.py:151` covers only None and the positive match — add not-found and `--col`-wins cases.

**4. Split the lint/whitespace churn into its own PR** — `pyproject.toml` (`"W"`), `ci.yml:10`, `.gitignore`, ~51 whitespace-only files incl. `anki_cli/db/anki_direct.py`. Unrelated to #33's scope, doubles the reviewable file count, and is a guaranteed conflict with #40 on `anki_direct.py`. Land it as `style:` (before or after #40), rebase this on it. Keep the SIM118 `dict(row)` fixes with the style PR.

**5. One `refactor:` commit carries ≥8 breaking changes; four are undisclosed** — `cli/commands/config.py:60-64`, `config_runtime.py:328-331`, `detect.py:132-136`. Disclosed: `--quiet/--verbose`, `init`, `status.message`, `--backend standalone`, exit 3. Not disclosed: `config:path` JSON drops `backups`/`standalone_default` (agents parse this); existing `[backup]`/`[review]`/`day_boundary_hour` are silently dropped from the TOML on next `config:set`; `prefer = "standalone"` / `ANKI_CLI_BACKEND=standalone` → exit 2 on every command with no migration hint; multi-profile users may silently switch collections. Please split into `style:` / `refactor:` / `feat(config)!:` / `refactor!:` commits with a `BREAKING CHANGE:` footer and add a "Breaking changes" section to the body.

**6. `anki_profile` plumbing is invisible to the suite** — `detect.py:50,57,72,77`, `app.py:131`, `tests/backends/test_detect.py:24`, `tests/cli/test_app_bootstrap.py`. The resolver stub is `lambda col_override, anki_profile=None: …` and the bootstrap stubs are `lambda **kwargs: …`. Deleting `anki_profile=anki_profile` at all five call sites → 539/539 green. Capture kwargs in the stubs and assert the value arrives.

**7. No test for backend-free commands under detection failure** — `test_app_bootstrap.py:96` invokes `main` with no subcommand. Add:

```python
@pytest.mark.parametrize("argv", [["version"], ["status"], ["config:path"]])
def test_backend_free_commands_survive_detection_failure(monkeypatch, argv):
    monkeypatch.setattr(app_mod, "detect_backend", _raise_exit3)
    assert CliRunner().invoke(app_mod.main, ["--format", "json", *argv]).exit_code == 0
```

**8. Docs drift** — `README.md:210-216, 83`, `SKILL.md:240-253`. Exit code 3 is now a real auto-mode outcome and is in neither exit-code table; `[collection] anki_profile` is now a live key and is documented nowhere. (Grep confirms no lingering `standalone`/`init`/`--quiet`/`backup`/`show_timer` references — good.)

## 🟡 Nits

- `config_runtime.py:301-305` — the `int` branch of `_coerce_raw_value` is dead by the same criterion used to delete list/dict (no int field remains); the PR swapped a public-API test for a private-function test to keep it covered. Delete it or comment why it stays.
- `config_runtime.py:14,166-169` — `prefer = "standalone"` locks out even `config:set` (exit 2). Map `standalone → auto` with a warning, or mention `--backend auto` in the error.
- `tui/_utils.py:8` — `_relative_eta` is now shared across modules; drop the underscore.
- `detect.py:132-137` — when a profile is configured, include it in `DetectionResult.reason` so `status` shows which one won.

## 💡 Suggestions

- `tui/_utils.py:8-20` — `_relative_eta` had no tests in either original copy and still has none; `days = hours // 24` (dropping the ceiling) passes. An 8-case parametrized test with `time.time` monkeypatched would pin it.
- `config_runtime.py:316` — the one *reachable* non-scalar rejection (key names a section) is untested; `config:set --key display --value '[1]'` with L316 replaced by `return raw_value` passes.
- `test_config_runtime.py` round-trip — `sections = list(data)[1:]` passes because reload refills defaults. Assert the written text contains all three section headers.

## ✅ What's good

- Dependency direction is right: `DEFAULT_ANKICONNECT_URL` and `JSONValue` moved into `models/`, so `backends → models`, not the reverse. Verified no import cycle.
- Test deletions reconcile exactly: 12 removed tests each target a deleted symbol; `test_set_config_value_rejects_unknown_section` is a proper tombstone.
- `test_auto_raises_when_nothing_found` pins both `exit_code == 3` and the message.
- SIM118 fixes are semantically exact — all ten use `dict(row)`, none use `for k in row` (which iterates *values* on `sqlite3.Row`).
- `sections = list(data)` removes a hardcoded list that would have silently dropped any future config section.

## Suggested path

Both blocking items stem from eager detection in `app.py`; the durable fix is #32 (lazy detection). If you'd rather, #32 could land first and this PR rebase on it, which would also make the `_BACKENDLESS` allowlist unnecessary. Either way: fix 1 and 3 (they're the same `models/config.py` change), split out the whitespace PR, and add the three tests in 6/7. Happy to re-review quickly once that's up.

ubermenchh:
This was merged by mistake while changes were still requested. Reverted in #45; the work is reopened for review as #46 with your commit and authorship intact. Sorry for the noise — please continue there.

Issue #30, opened 2026-09-12
Backend Protocol parity: normalize return shapes, lift _store methods, align search semantics


ubermenchh:
## Problem

`backends/protocol.py` holds for **signatures** but not for **return shapes** or **query semantics**, so callers know which backend they're on.

### Return shapes differ
- `get_card` / `get_note`: Direct (`db/anki_direct.py:609-756`) emits `{cardId, note, deckName, notetype_name, fields: list, due_info…}`; AnkiConnect (`backends/ankiconnect.py:555-591`) passes raw `cardsInfo`/`notesInfo` (`modelName`, `fields: {name: {value, order}}`, `noteId`).
- `get_deck`: Direct adds `kind/stats/next_due`; AC only `id/name/due_counts`.
- Consumers compensate with dual-shape branches: `cli/commands/review.py:107-113`, `cli/commands/cards.py:145-162`, `core/scheduler.py:43-45`, and the `_extract_note_id`-style key probing.

### Methods missing from the Protocol
`preview_ratings`, `snapshot_card_state`, `restore_card_state`, `get_next_due_card` live only on `DirectBackend._store`. Callers probe `backend.name == "direct"` / `hasattr(backend, "_store")` in ~9 places: `cli/commands/review.py:208-215, 300-303, 327-343, 380-384`, `tui/review_app.py:459, 691`, `tui/repl.py:362, 442, 479`.

### Query semantics differ for the same string
- `is:due` includes `queue = 0` (new) on Direct (`core/search.py:637-645`) but not on AnkiConnect/Anki. So `get_due_counts` on AC (`ankiconnect.py:764-775`) reports `new: 0` always, and `review`, `deck show`, `pick_next_due_card_id` (`core/scheduler.py:28-30`) inherit the mismatch.
- `deck:Foo` → `LIKE 'Foo'` on Direct (`search.py:485-490`); Anki also matches `Foo::Bar`. Same for `_deck_filter` at `anki_direct.py:2124` (`review --deck Japanese` ignores `Japanese::Core`).
- `tag:foo` doesn't match `foo::child` (`search.py:499-505`).
- `added:N` filters on `n.mod` (modification) (`search.py:513-519`); Anki uses creation time (`n.id / 1000`).
- Unknown prefixes (`card:`, `note:`, `rated:`, `mid:`, `did:`, `dupe:`) fall through to full-text search on Direct (`search.py:320-322`) but are real filters on AC.

### `answer_card` means different things
[...]

PR #48, opened 2026-09-12
fix(direct): treat day-learn (queue 3) due as a day index, not an epoch

### LINKED ISSUE #19: queue=3 (day-learn) treated as epoch seconds -> wrong due, 1970 dates on unsuspend
## Problem

In Anki, `queue = 1` is intraday learning (`due` = epoch seconds) but `queue = 3` is **day-learn** (`due` = day index relative to `col.crt`), produced by any learning step ≥ 1 day (e.g. steps `10m 1d`).

`anki_cli/db/anki_direct.py` treats both as epoch seconds:

- `:773`, `:803` — `get_due_counts` uses `due <= now_sec` for `queue IN (1, 3)`.
- `:2547-2548` — `_card_row_to_fsrs` does `fromtimestamp(due_raw)` for `queue in (1, 3)`.
- `:2763` — labels it `learn_epoch_secs`.
- `:1091-1096`, `:2397-2402` — unsuspend/unbury `CASE` maps `type = 1` → `queue = 1` unconditionally.

## Reproduce

Card `type=1, queue=3, due=20050` (30 days out).

- `get_due_counts` counts it as due now.
- `get_next_due_card` returns it first.
- `_card_row_to_fsrs` yields `due = 1970-01-01T05:34:10Z`.
- `card:suspend` then `card:unsuspend` → `queue=1, due=20050` → due in 1970, shown immediately and forever.

## Suggested fix

- Treat `queue == 3` as a day index everywhere: `due <= today_days`, epoch = `(crt_day + due) * 86400`.
- Restore queue with Anki's rule:
  ```sql
  CASE WHEN type IN (1, 3) THEN (CASE WHEN due > 1000000000 THEN 1 ELSE 3 END) ...
  ```
- Consolidate the three copies of the `queue = CASE …` SQL into one helper.

Add tests with a `queue = 3` card for due counts, next-due, and unsuspend.

### LINKED ISSUE #43: Queue-from-type CASE exists in four places; three map type 3 -> queue 3 unconditionally
`_return_cards_from_filtered_decks` (added in #40) computes queue from type correctly, splitting types 1/3 on `due > 1000000000` (intraday epoch vs day index):

```sql
CASE
  WHEN queue < 0 THEN queue
  WHEN type = 0 THEN 0
  WHEN type IN (1, 3) THEN CASE WHEN due > 1000000000 THEN 1 ELSE 3 END
  ELSE 2
END
```

The three pre-existing copies in `unsuspend_cards`, `unbury_cards`, and `reset_cards` (`anki_cli/db/anki_direct.py`, grep `queue = CASE`) map `type = 3` to `queue = 3` unconditionally, which is wrong for intraday relearning (a relearning card with an epoch `due` becomes a day-learn card due in 1970). Overlaps with #19.

Fix: hoist the correct CASE into a module-level constant (`_QUEUE_FROM_TYPE_SQL`) and use it in all four sites. One parametrized integration test over (type, due, expected queue).

ubermenchh:
Closes #19
Closes #43

## Problem

Anki stores two units in `cards.due` for learning cards: intraday learning (**queue 1**) holds a unix epoch; day-learn (**queue 3**, produced by any learning step ≥ 1 day) holds a **day index** relative to `col.crt`. The direct backend read both as epochs, so a day-learn card 30 days out:

- counted as due now (`get_due_counts`), sorted first (`get_next_due_card`),
- was fed to FSRS with a 1970 due date,
- came back from `suspend` → `unsuspend` as queue 1 with a day-index due → due in 1970 forever.

## Changes

**One rule, one place.** `queue_from_type_sql()` mirrors rslib's `restore_queue_from_type` (`type 1/3 → queue 1 if due > 1_000_000_000 else 3`) and replaces four hand-written CASEs (unsuspend, two unbury copies, filtered-deck return). Three of those mapped `type 3 → queue 3` unconditionally, which was wrong for intraday relearning (#43).

**Read side.** `get_due_counts`, `get_next_due_card`, `_get_next_due_for_deck` compare queue 3 against today's day index and order learning cards by absolute time across units. `_get_next_due_for_deck` also fixes the pre-existing `due * 86400` ordering (compared a day index against an epoch; item 6 in #22).

**Decode.** `_decode_due` emits a new `kind: "learn_day_index"` (with `day_index` + `epoch_secs`) for day-learn cards; `learn_epoch_secs` is unchanged for intraday cards. TUI due formatters treat the new kind like a review day index.

**FSRS.** `_card_row_to_fsrs` converts day-learn due via `col.crt`, and `Relearning` is now decided by `type == 3` rather than `queue == 3` (day-learn is a queue, not a state). `_map_fsrs_result_to_anki` writes queue 3 with a day-index due when the next step is ≥ 1 day, as Anki v3 does, so what we write round-trips through the restore rule.

## Tests

New `tests/integration/test_anki_direct_learn_queue_units.py` (13 tests): the threshold, the SQL CASE against rslib's truth table, decode for both units (including a suspended relearning card), both FSRS directions with a pinned clock, and next-due ordering across units. Unsuspend/unbury tests now use realistic due values for both learn encodings (the old ones encoded the wrong `type 3 → queue 3` rule). Due-count/next-due tests gain a far-future day-learn card that must not be counted as due.

636 + 57 tui pass. Sanity-checked on a copy of a real collection (17 intraday learning cards): counts unchanged, suspend/unsuspend round-trips a learning card to the same `(queue, due)`.

## Not in this PR

- `is:due` still includes `queue = 0` (Anki excludes new) → #30.
- Rollover hour / local-midnight day index → #21.


ubermenchh:
Review follow-ups landed in the second commit:

**Important**
- `is:due` in `core/search.py` still compared queue 3 against `now_sec` — fixed (`queue IN (2, 3)` vs day index). The `find_queries` fixture had a queue-3 card with an epoch due, i.e. the impossible encoding; corrected, plus a far-future day-learn card that must not match.
- `queue_from_type_sql` now defaults to `RESTORED_DUE_SQL` (`odue` when set), so unsuspend/unbury of a card parked in a filtered deck read the real scheduling due, matching rslib `restore_queue_after_bury_or_suspend`. Tests for both.
- `_map_fsrs_result_to_anki` takes `now_dt` (the `review_datetime` handed to FSRS) rather than re-reading the clock; tests inject it instead of monkeypatching `datetime`. A test pins the 0.999s-boundary case.

**Suggestion taken**: day-learn due is now `today + round(step / 86400)` per rslib `InSecs → InDays`, parametrized over 24/36/47/60/61h.

**Nit**: revlog logs day-learn `ivl`/`lastIvl` in positive days; a real-path `answer_card` test (no `_map_fsrs` stub) covers it.

Not taken: `scheduler.py` comparing raw `due` across queues in `pick_next_due_card_id` — that's the AnkiConnect-shared path and belongs with #30's shape normalization; noted there.

645 + 57 tui pass.

Issue #1, opened 2026-02-22
Scrolling through the cards


ilya-grigoriev:
It would be cool to be able to scroll (vertically and horizontally) through the long list of the cards when searching.

ilya-grigoriev:
Awesome!

Issue #10, opened 2026-03-10
Update the TUI


ubermenchh:
- The idea is to have a full standalone tui along with the shell.
- I generated these designs for inspiration.

<img width="1377" height="1070" alt="Image" src="https://github.com/user-attachments/assets/ddee123a-0246-4881-ad7e-eebc31d0f8a3" />

<img width="1395" height="1241" alt="Image" src="https://github.com/user-attachments/assets/536551b6-9a78-411a-9538-3fcebf4b6cbb" />

<img width="1396" height="841" alt="Image" src="https://github.com/user-attachments/assets/253a611e-327d-4ecd-9f68-acd9fcdcc251" />

umangk-2410:
Concrete layout plan for this is now in #84 (redesign) with the measured layout bugs split out in #78–#83; tracked in #88.

Issue #3, opened 2026-02-25
Front and Back of cards is not shown with AnkiConnect as backend


ilya-grigoriev:
<img width="1919" height="470" alt="Image" src="https://github.com/user-attachments/assets/e6b5de37-19f1-4689-ad31-c436afa001d2" />

PR #49, opened 2026-09-13
fix: TUI hint bars, literal newlines in rating buttons, and undo edge cases

### LINKED ISSUE #34: TUI bugs: literal '\n' in rating buttons, phantom hint keys, undo stack/revlog/undo.json issues
Small, verified TUI bugs.

### 1. Rating buttons render a literal `\n`
`tui/review_app.py:613-616`:
```python
self._ui_update("#rate-1", f"1 Again\\n{hints[1]}")
```
`\\n` inside an f-string is the two characters `\` `n`, not a newline. Should be `\n`.

### 2. Hint bars advertise keys that don't exist
- `review_app.py:620-622` shows `e edit`, `m mark`, `s suspend`; `BINDINGS` (`:359-371`) define none of them.
- `browse_app.py:817-818` shows `a add`; no binding.
- Browse preview says `Enter Study` (`:844, :904`) but Enter opens the detail modal (`:535`).
- `browse_app.py:644-650` `action_edit_selected` is a stub that says "not implemented".

Fix: derive the hint bar from `self.BINDINGS` (those with `show=True`) so it can't drift. Then either implement `e`/`s`/`m`/`a` (issue #10 "Update the TUI" is the place) or remove them from the bar.

### 3. Undo stack gets a stale entry on failed answer
`review_app.py:463-477` pushes the undo snapshot, *then* calls `answer_card`, which can raise (e.g. `_ensure_write_safe` RuntimeError because Anki was launched after the card loaded). The exception is caught at `:475` but the entry is never popped, so the next `u` is a no-op "undone".

Fix: push after success, or `self._undo.pop()` in the `except`.

### 4. `restore_card_state` appends instead of deleting the undone revlog row
`db/anki_direct.py:920-928` appends a `type=4` revlog row and keeps the original review row (Anki's undo deletes it). For legacy cards seeded from revlog, the next `answer` computes a different stability than before the undone review.

Fix: `DELETE FROM revlog WHERE cid = ? AND id > ?` using `snapshot["created_at_epoch_ms"]`.

### 5. `undo.json` is not written atomically
`core/undo.py:88` — `write_text` directly. Use `NamedTemporaryFile` in the same dir + `os.replace`.

LouisDeconinck:
Closes #34

Fixes each item in the issue:

### 1. Literal `\n` in rating buttons
`tui/review_app.py` — `_refresh_rate_buttons` used `\\n` inside f-strings, rendering the two characters \\n instead of a newline. Now uses `\\n` (a real newline).

### 2. Hint bars advertised keys that don't exist
- `review_app.py`: `_render_hint_bar` is now derived from `BINDINGS` (entries with `show=True`), so it can only advertise keys that actually do something. This removes the phantom `e edit`, `m mark`, and `s suspend` hints; the side-panel SHORTCUTS block drops `e edit` too. `escape` (blur command) stays bound but hidden via `show=False`.
- `browse_app.py`: same derivation. `a add` is gone (never bound); `e`/`action_edit_selected` was a stub that just opened the detail modal with a "not implemented" status, so the binding and the stub action are removed (real editing is tracked in #10). Preview action bars now label Enter as `Detail` instead of `Study` (Enter opens the detail modal) and no longer show `e Edit`.

### 3. Stale undo entry on failed answer
`action_rate` pushed the snapshot before `answer_card`; when the answer raised (e.g. write-safety), the stale entry made the next `u` a no-op "undone". The snapshot is still taken before the answer but pushed only after it succeeds — and the post-commit push is best-effort (`try/except OSError` + warning) so a failed undo write cannot fail a command whose answer is already durable. Same fix applied to the identical pattern in `tui/repl.py` and `cli/commands/review.py` (`review:answer`).

### 4. `restore_card_state` appended instead of deleting the undone revlog row
`db/anki_direct.py`: `answer_card` now returns the `revlog_id` of the row it wrote; the three undo-push sites merge it into the snapshot (`snapshot={**snap, "revlog_id": result["revlog_id"]}`), and `restore_card_state` deletes exactly that row (`DELETE FROM revlog WHERE id = ? AND cid = ? AND usn = -1`) instead of inserting a `type=4` row — matching Anki's undo. Correction to the original rationale: the stale-stability cause was the undone review's *own* `ease∈1..4` row surviving the undo; the old compensating `type=4, ease=0` row was already ignored by `_seed_fsrs_card_from_revlog`. The `usn = -1` guard never touches a row that already synced, since revlog deletions don't propagate to AnkiWeb. `snapshot_card_state` is a pure cards-row projection again (no hidden timestamp field). Snapshots without `revlog_id` (e.g. pre-fix `undo.json` entries) leave revlog untouched.

JSON contract notes: `review:undo` reports `revlog_deleted` instead of `revlog_id`, and `review:answer` output gains `revlog_id` (direct backend only).

### 5. Non-atomic undo.json writes
`core/undo.py`: `_save` writes via `NamedTemporaryFile` in the same directory + `os.replace`, with temp-file cleanup on failure (including a `tmp.write` failure).

### Tests
- `test_anki_direct_snapshot_restore.py`: snapshot is a pure row projection; restore deletes only the recorded `revlog_id` row (other unsynced rows, synced `usn != -1` rows, other cards' rows, and non-int ids all survive); plus a real-clock `snapshot → answer_card → restore` round trip.
- `test_undo.py`: failure-path tests proving a failed `os.replace` *and* a failed `tmp.write` keep the existing file and leave no temp files.
- `test_review_commands.py`: direct-backend `review:answer` failure leaves no undo entry; an `OSError` from `UndoStore.push` still exits 0 with a stderr warning.
- `test_review_app.py` / `test_browse_app.py`: real newlines in rate buttons, exact hint-bar text derived from `BINDINGS`, the removed edit stub, `Enter Detail` labels, snapshot-before-answer ordering, and that failed/push-failed/successful answers behave correctly.

`ruff check anki_cli`, `ty check anki_cli`, and `pytest` (655 passed) + `pytest -m tui` (67 passed) all green.


ubermenchh (changes requested):
# Code Review — TUI hint bars, literal newlines, undo edge cases

**Verdict:** REQUEST CHANGES
**Counts:** 🟠 3 important · 🟡 7 nit · 💡 6 suggestion · ✅ 5 praise

## Summary

Fixes 1, 2, 3 and 5 (real `\n`, BINDINGS-derived hint bars, push-undo-after-success, atomic `undo.json`) are correct, small, and well tested. Fix 4 (`DELETE FROM revlog WHERE cid=? AND id>?`) is the concern: keying the deletion on a wall-clock window instead of the specific revlog row the review wrote means a persisted undo entry can later delete revlog rows written by *other* reviews (including already-synced ones from Anki Desktop), which is irreversible and never propagates to AnkiWeb. Fix 4 also lacks tests on its boundary and on the CLI/REPL surfaces.

Recommended next step: delete by `revlog_id` (returned from `answer_card`) or at minimum add `AND usn = -1`, wrap the post-commit `push` in a try/except, and add the missing CLI failure test.

One factual correction to the PR description: the old `type=4, ease=0` row was already ignored by `_seed_fsrs_card_from_revlog` (`anki_direct.py:3194`, `ease IN (1,2,3,4)`). The stale-stability cause was the undone review's *own* `ease∈1..4` row surviving. The fix is still right; the rationale in the comment at `anki_direct.py:1114-1117` is slightly off.

## 🟠 Important

### 1. Revlog deletion keyed on a time window can delete rows the undone review did not write
**File:** `anki_cli/db/anki_direct.py:1065`, `:1118-1124`

`UndoStore` is persistent and collection-scoped, not session-scoped (`core/undo.py:47-75`). Trace: answer card 100 via CLI at T1 → revlog `id=T1', usn=-1`; entry pushed. Later the user reviews card 100 in Anki Desktop at T3 and syncs (both rows now `usn=N`). Next day `anki review:undo` pops the T1 entry → `DELETE ... id > T1` removes **both** rows (`revlog_deleted=2`). Anki's sync has no revlog graves, so the Desktop row is gone locally, survives on AnkiWeb, and FSRS seeding/stats diverge per device until a one-way full sync. Base did not delete anything, so this is a new failure class. Second trigger: a revlog row synced from a device with a wrong clock (`id` in the future) is deleted by *any* undo of that card.

This also introduces two timestamps for one event (`UndoItem.created_at_epoch_ms` set by callers, plus the new hidden one inside the snapshot dict at L1065), and turns `snapshot_card_state`'s return value from a pure row projection into projection-plus-control-field — the test had to `snap.pop("created_at_epoch_ms")` to keep asserting the shape.

**Fix (recommended):** delete exactly the row that was written. `answer_card` already has `revlog_id` at L2331:
```python
# answer_card return dict (L2374)
return {..., "revlog_id": revlog_id}

# callers: push happens after answer_card now, so the id is available
undo_item = UndoItem(..., snapshot={**snap, "revlog_id": result["revlog_id"]})

# restore_card_state
rid = snapshot.get("revlog_id")
if isinstance(rid, int) and not isinstance(rid, bool):
    revlog_deleted = conn.execute(
        "DELETE FROM revlog WHERE id = ? AND cid = ? AND usn = -1",
        (rid, card_id),
    ).rowcount
```
**Fix (minimum):** keep the window, but never touch synced rows and drop the duplicate clock:
```python
def restore_card_state(self, snapshot, *, since_epoch_ms: int | None = None):
    ...
    if since_epoch_ms is not None:
        revlog_deleted = conn.execute(
            "DELETE FROM revlog WHERE cid = ? AND id >= ? AND usn = -1",
            (card_id, since_epoch_ms),
        ).rowcount
```
and pass `since_epoch_ms=item.created_at_epoch_ms` at the three restore sites (`review.py:343`, `review_app.py:501`, `repl.py:456`). Remove L1065. Bonus: pre-PR `undo.json` entries already carry `created_at_epoch_ms` on the item, so they'd get cleanup too.

### 2. Moving `push` after the commit turns an undo-store write failure into a failed command with a durable side effect
**File:** `anki_cli/cli/commands/review.py:398-399` (same at `tui/review_app.py:480-481`, `tui/repl.py:505-506`)

Trace: `~/.local/share/anki-cli/` unwritable or ENOSPC. `answer_card` commits (`_connect_write` L71-79). `UndoStore().push()` → `_save` raises `OSError`. Not in the except tuples at L400/L402 → traceback, exit 1, `emit_success` never runs. A script consumer retries → second revlog row, card rescheduled off two reviews. On base the same misconfiguration failed *before* the answer, so retry was idempotent. In `review_app` the exception escapes a Textual action after the DB commit; in `repl` it escapes the loop.

**Fix:** undo persistence is best-effort once the answer is committed:
```python
result = backend.answer_card(card_id=int(card_id), ease=ease)
if undo_item is not None:
    try:
        UndoStore().push(undo_item)
    except OSError as exc:
        click.echo(f"warning: answer saved but undo entry not written: {exc}", err=True)
```
Mirror with `self._set_status("answered; undo not saved: ...")` in `review_app` and a `console.print` in `repl`.

### 3. Fix 3 is regression-guarded at one of three sites; fix 4 has no boundary or round-trip test
**File:** `tests/cli/test_review_commands.py:465-497`, `anki_cli/tui/repl.py:478-511`, `tests/integration/test_anki_direct_snapshot_restore.py:145-192`

- CLI: the only `review:answer` failure test uses `name = "ankiconnect"`, so the `undo_item` branch never runs. Reverting `review.py` to base keeps the suite green.
- REPL: `_inline_review` has zero tests.
- Revlog: rows sit at 1000/5000/6000 with snapshot ts 2000, so `>` vs `>=` is not distinguished. `created_at > 0` also survives a mutation to `int(time.time())` (seconds), which would make restore run `id > 1.7e9` and wipe the card's entire history. No test runs `snapshot → answer_card → restore` with real clocks.

**Fix:** add (a) a direct-backend `review:answer` test where `answer_card` raises `LookupError`, asserting exit 1 and `pushed == []`; (b) a round-trip integration test:
```python
snap = store.snapshot_card_state(100)
store.answer_card(100, ease=3)
assert len(_revlog_rows(db_path)) == 1
result = store.restore_card_state(snap)
assert result["revlog_deleted"] == 1 and _revlog_rows(db_path) == []
```
(c) bracket the timestamp: `before <= snap.pop("created_at_epoch_ms") <= after` with both taken as `int(time.time()*1000)`. A REPL test is doable (everything is monkeypatchable) but acceptable as follow-up.

## 🟡 Nits

- **`_save` leaks the temp file if `tmp.write` raises** — `core/undo.py:106-107`. `tmp_name = tmp.name` is assigned *after* `tmp.write(payload)`; on ENOSPC the `finally` sees `""` and skips unlink. Swap the two lines. The new test only injects at `os.replace`, so it doesn't cover this.
- **`id > ?` boundary is unpinned** — `anki_direct.py:1122`. Same-ms equality is unlikely in practice (a `pgrep` subprocess in `_ensure_write_safe` sits between snapshot and answer), but nothing pins the choice. Add a revlog row at `id == created_at_epoch_ms` to the test and use `>=` (snapshot precedes the answer, so equality belongs to the undone review). Moot if you adopt delete-by-`revlog_id`.
- **Harden the timestamp check** — `anki_direct.py:1120`. `isinstance(True, int)` is `True`; `"created_at_epoch_ms": true` → `id > 1` → deletes the card's whole history. No production writer emits a bool, so defensive only: `isinstance(v, int) and not isinstance(v, bool) and v > 946_684_800_000`.
- **`revlog_id` → `revlog_deleted` is a `review:undo` JSON contract change** — `anki_direct.py:1126-1130` emitted verbatim by `review.py:354`. Undocumented field with no consumers found in repo/README/SKILL.md, so fine, but worth calling out in the PR description.
- **Browse hint-bar test has no negative branch** — `tests/tui/test_browse_app.py:575-589`. Deleting `if not binding.show: continue` still passes (phantoms "add"/"edit" were never bindings). `q`/`escape` share "Quit" so a per-binding `else` won't work; assert the exact `.plain` string instead, which also catches key/label swaps and dropped separators. Same recommendation for the review_app test.
- **`test_action_rate_successful_answer_pushes_undo` doesn't pin snapshot-before-answer** — `tests/tui/test_review_app.py:706-733`. The fake store returns a constant, so moving `snapshot_card_state` below `answer_card` still passes. Make `answer_card` mutate `Store.queue` and assert the snapshot has the pre-answer value. Also add `assert app._answered == 0` and a `_load_next` counter to the failure test.
- **Pre-existing: exception text goes through Rich markup** — `tui/repl.py:501` (moved, not new; also `:461`, `review_app.py:477`). A message containing `[/x]` raises `MarkupError` inside the except handler. `from rich.markup import escape` → `escape(msg)`. Not this PR's bug.

## 💡 Suggestions

- **Split the revlog-deletion change into its own commit/PR.** Fixes 1/2/3/5 are zero-risk; fix 4 irreversibly mutates collection data and inverts a documented invariant (`# we never delete revlog`). Reverting one currently drags the others.
- **`answer_with_undo(backend, undo, *, card_id, ease)` helper in `core/undo.py`.** The ~20-line snapshot→answer→push block is now hand-rolled identically at `review_app.py:458-481`, `repl.py:478-506`, `review.py:379-399`; this PR had to apply the same fix three times, and only one copy has a test. The mirror pop→restore block is also triplicated.
- **Add `snapshot_card_state` / `restore_card_state` / `preview_ratings` pass-throughs to `DirectBackend`** (`backends/direct.py`). Removes 16 `cast(Any, backend._store)` sites in TUI/CLI and lets the `name == "direct" and hasattr(_store)` checks become a capability check.
- **`hint_bar(bindings) -> Text` pure function.** `_render_hint_bar` is now byte-identical in both apps (`review_app.py:622-635`, `browse_app.py:805-818`); the "key chip" pattern also appears in `_render_shortcuts` and both preview-action renderers. The two preview-action lists in `browse_app.py:834/:894` should at least be one method with a `suspend_label` param.
- **`fsync` before `os.replace`** — `core/undo.py:106`. The comment claims crash safety; that's true for process crashes, not OS crashes. Low priority for an undo cache (`_load` already tolerates garbage).
- **Document the semantics change.** README:189 and SKILL.md:316 describe `review:undo` only as "undo last answer". One sentence: "Undo removes the revlog rows written by the undone review (matching Anki's own undo)." Drive-by for a separate PR: `_allocate_row_id` and `_allocate_epoch_ms_id` (`anki_direct.py:2531-2547`) are byte-identical.

## ✅ Praise

- **Hint bar derived from `BINDINGS`** (`review_app.py:622-635`, `browse_app.py:805-818`). Deletes the second source of truth instead of syncing it; `show=False` on `escape` is exactly Textual's mechanism for bound-but-unadvertised. Removing the `action_edit_selected` stub rather than keeping it "for later" is the right call. Verified neither app mounts `Footer`, so `show=True` flips have no other UI surface.
- **Atomic `_save` with a test that checks both invariants** (`core/undo.py:96-112`, `tests/unit/test_undo.py:755-771`). Same-dir temp + `os.replace` + `finally` cleanup is the minimal correct shape; `NamedTemporaryFile` also makes `undo.json` 0600 instead of umask-dependent. The test asserts on filesystem state (original bytes intact, directory listing clean), not on mocks.
- **`test_action_rate_failed_answer_leaves_no_undo_entry` genuinely fails on base** — push preceded the raising `answer_card`, so `pop()` returned the item. Real regression guard.
- **`test_restore_card_state_without_snapshot_timestamp_leaves_revlog`** pins backward compatibility with pre-PR `undo.json` entries — the non-obvious case.
- **New `DELETE` is parameterized and inside `_connect_write`** (`anki_direct.py:1075`), so it's gated by `_ensure_write_safe` and shares the `BEGIN IMMEDIATE ... COMMIT` with the card `UPDATE` — a failure rolls back both.

---
*Review limits: I did not run the test suite locally; reviewed from the PR head via `git show`. The sync-divergence claim in Important #1 rests on Anki's sync protocol having no revlog graves — worth a quick check against the Anki source before restructuring.*

LouisDeconinck:

ubermenchh (changes requested):
# Re-review after force-push (`121f205` → `53d8556`)

**Verdict:** REQUEST CHANGES (unchanged)

## What changed

The new head is a rebase onto `main` (picking up #48) plus a trailing-newline fix in `tests/tui/test_browse_app.py`. I diffed the PR's own changes against their respective bases: the source-side diff is **byte-identical** to the previously reviewed revision, and #48 does not touch `snapshot_card_state`, `restore_card_state`, or revlog id allocation, so the rebase is clean and does not interact with any finding.

Verified in an isolated worktree at `53d8556`: `pytest` 647 passed / 4 skipped, `pytest -m tui` 66 passed, `ruff check` clean. Matches your comment.

## Status of prior findings

All three Important findings are still open — none of the code involved was modified:

| # | Finding | Location | Status |
|---|---|---|---|
| 1 | Revlog `DELETE` keyed on a wall-clock window can remove rows written by other/synced reviews; duplicate timestamp inside the snapshot dict | `anki_direct.py:1105`, `:1158-1164` | Open |
| 2 | `UndoStore.push` after the DB commit raises uncaught `OSError` → failed exit with a durable review; retry double-answers | `review.py:398-399`, `review_app.py:480-481`, `repl.py:505-506` | Open |
| 3 | Missing CLI direct-backend failure test, no REPL test, no revlog boundary / round-trip test | `tests/cli/test_review_commands.py`, `tests/integration/test_anki_direct_snapshot_restore.py` | Open |

Nits (temp-file leak when `tmp.write` raises at `core/undo.py:106-107`, `bool`/lower-bound guard on the timestamp, `revlog_id`→`revlog_deleted` contract note, browse hint-bar negative assertion, snapshot-before-answer pin) are likewise unchanged.

## Empirical confirmation of the coverage gaps (#3)

Since I had the worktree, I applied each mutation and ran the suite. All three survive on `53d8556`:

| Mutation | Result |
|---|---|
| `anki_direct.py`: `id > ?` → `id >= ?` | `test_anki_direct_snapshot_restore.py`: 6 passed |
| `anki_direct.py`: `int(time.time() * 1000)` → `int(time.time())` (seconds) — on a real collection this makes undo run `DELETE ... id > ~1.7e9`, wiping the card's whole revlog | `test_anki_direct_snapshot_restore.py`: 6 passed |
| `review.py`: move `UndoStore().push(undo_item)` back above `answer_card` (i.e. revert fix 3 at the CLI site) | `tests/cli`: 246 passed |

The second one is the reason I'd like at least the round-trip test (`snapshot → answer_card → restore`, asserting `revlog_deleted == 1` and the row is gone) and the `before <= ts <= after` bound before this merges, independent of how #1 is resolved.

## Suggested minimal path to approval

1. Resolve #1 either by deleting by `revlog_id` (returned from `answer_card`, carried in the undo item) or, at minimum, `DELETE ... WHERE cid = ? AND id >= ? AND usn = -1` with the cutoff passed as `since_epoch_ms=item.created_at_epoch_ms` instead of stored inside the snapshot.
2. Wrap the three post-commit `push` calls in `try/except OSError` and surface a warning instead of failing the command.
3. Add the CLI direct-failure test, the round-trip integration test, and the timestamp bound.
4. Swap `tmp_name = tmp.name` above `tmp.write(payload)` in `core/undo.py`.

Happy to re-review promptly once those land.

LouisDeconinck:

ubermenchh (changes requested):
# Re-review after second force-push (`53d8556` → `438f31c`)

**Verdict:** REQUEST CHANGES (unchanged)

## What changed

Rebase onto `1cbaf6d` (#51 style pass) plus two more trailing-newline fixes in tests. The PR's own source diff against its parent is again **byte-identical** to the revision reviewed twice before; no finding has been addressed yet.

Note `main` has moved again since your push: #53 (`125d113`, filtered-deck `odid`/`odue`) landed at 09:28 UTC and edits `snapshot_card_state` and `restore_card_state` — the same functions this PR touches. GitHub reports the PR as mergeable, and I confirmed it in a worktree: a test merge of `438f31c` with `origin/main` is conflict-free and the merged tree passes `pytest` (657) + `pytest -m tui` (66) + `ruff`. So no action needed on the rebase, but the snapshot dict now carries both #53's `odid`/`odue` and this PR's `created_at_epoch_ms`; Important #1's suggestion to move the cutoff out of the snapshot dict is a little more relevant now that the dict has two owners.

## Status of prior findings

| # | Finding | Status |
|---|---|---|
| 1 | Time-window revlog `DELETE` can remove rows from other/synced reviews; cutoff hidden inside the snapshot dict | Open |
| 2 | Post-commit `UndoStore.push` raises uncaught `OSError` → command fails after the review is durable | Open |
| 3 | No CLI direct-failure test, no REPL test, no revlog boundary/round-trip test | Open |
| — | All nits (temp-file leak order in `_save`, `bool`/lower-bound guard, `revlog_id`→`revlog_deleted` note, browse hint-bar negative assertion, snapshot-before-answer pin, pre-existing Rich markup escape) | Open |

## New: the `id > ?` boundary is a demonstrated bug, not a theoretical one

I previously filed the `>` vs `>=` question as a nit because a `pgrep` subprocess inside `_ensure_write_safe` sits between the snapshot and the revlog insert in production. I ran a real-clock round-trip probe (`snapshot_card_state → answer_card → restore_card_state`, no id mocking) on the merged tree, 30 iterations each:

| Setup | Result |
|---|---|
| Real `_ensure_write_safe` (production path on macOS, ~60ms `pgrep`) | 30/30 delete the review's revlog row |
| `_ensure_write_safe` stubbed — **which every existing integration test in the repo does** | **7/30 leave the row in place**, all with `revlog_id == created_at_epoch_ms` (delta 0 ms) |

So the only thing preventing `restore_card_state` from missing the undone review's revlog row is the incidental latency of a process scan — which is faster on Linux (`/proc` walk), may be short-circuited in future, and is absent in tests. This is exactly the stale-FSRS-seed symptom the PR sets out to fix. I'm upgrading this from nit to **Important**; the remedy is the same as #1 (delete by `revlog_id`, or `id >= ?` with `AND usn = -1`).

Probe for reference (drop into `tests/integration/`, uses the helpers already in `test_anki_direct_answer_card.py`):
```python
def test_restore_deletes_the_undone_reviews_revlog_row(monkeypatch, tmp_path):
    store, db_path = _make_store(tmp_path)
    monkeypatch.setattr(store, "_ensure_write_safe", lambda: None)
    _fake_scheduler(monkeypatch, store, {})

    snap = store.snapshot_card_state(100)
    store.answer_card(100, ease=3)
    assert len(_revlog_rows(db_path)) == 1

    result = store.restore_card_state(snap)

    assert result["revlog_deleted"] == 1
    assert _revlog_rows(db_path) == []
```
This fails ~25% of the time on `438f31c` and deterministically passes with `>=`.

Mutants from the last review (`>`→`>=`, ms→seconds, revert CLI push order) still all survive the suite.

## Minimal path to approval (unchanged)

1. Resolve #1: delete by `revlog_id` returned from `answer_card` and carried in the undo item; or at minimum `DELETE ... WHERE cid = ? AND id >= ? AND usn = -1` with the cutoff passed as `since_epoch_ms=item.created_at_epoch_ms` rather than stored inside the snapshot.
2. Wrap the three post-commit `push` calls in `try/except OSError` and surface a warning instead of failing the command.
3. Add the CLI direct-failure test, the round-trip test above, and a `before <= ts <= after` bound on the snapshot timestamp.
4. Swap `tmp_name = tmp.name` above `tmp.write(payload)` in `core/undo.py`.

Rebasing onto `125d113` is optional (clean merge), but if you do, please keep the PR's changes in a separate commit from the rebase so they can be diffed.

LouisDeconinck:
Thanks for the detailed re-reviews — all addressed in `63e5805` (new commit, no rebase; a `git merge-tree` check against `125d113` is still conflict-free and the merged `restore_card_state` keeps #53's `odid`/`odue` alongside the new delete).

## 🟠 Important

**1. Revlog deletion keyed on a time window** — adopted the recommended fix. `answer_card` now returns `revlog_id` (`anki_direct.py:2425`); each of the three push sites merges it into the snapshot (`snapshot={**snap, "revlog_id": result["revlog_id"]}`), and `restore_card_state` deletes exactly `DELETE FROM revlog WHERE id = ? AND cid = ? AND usn = -1` (`anki_direct.py:1162-1168`). `created_at_epoch_ms` is out of `snapshot_card_state` entirely — the snapshot is a pure row projection again, which also sidesteps the `>`/`>=` boundary and the ms-vs-seconds hazards (both mutants now have no code to attack). The `usn = -1` guard means a row that already synced is left in place rather than diverging from AnkiWeb; `restore` reports `revlog_deleted: 0` in that case.

**2. Post-commit `push` failure** — `UndoStore().push()` is wrapped in `try/except OSError` at all three sites: `review.py` emits `warning: answer saved but undo entry not written: ...` on stderr, `review_app` sets `answered; undo not saved: ...` status, `repl` prints the equivalent. Exit code / action flow are unaffected.

**3. Missing tests** — added:
- `test_review_answer_direct_failure_leaves_no_undo_entry` (direct backend, `answer_card` raises `LookupError` → exit 1, `pushed == []`). This now fails if the push is moved back above `answer_card`.
- `test_review_answer_direct_undo_push_failure_still_succeeds` (CLI) and `test_action_rate_undo_push_failure_still_counts_answer` (TUI) — pin finding #2.
- `test_restore_deletes_the_undone_reviews_revlog_row` — real-clock `snapshot → answer_card → restore` round trip (`_ensure_write_safe` stubbed, `_allocate_epoch_ms_id` left real): asserts the row `answer_card` wrote is exactly `result["revlog_id"]`, then `revlog_deleted == 1` and revlog is empty.
- Boundary cases: only the recorded row is deleted (a *later* unsynced row for the same card survives — your #1 scenario), synced `usn != -1` rows survive (incl. a round-trip where the row syncs before undo), `revlog_id` pointing at another card's row is ignored, and non-int/bool/float `revlog_id` values are ignored.
- The `before <= ts <= after` bound is moot now — the snapshot carries no timestamp; only `UndoItem.created_at_epoch_ms` remains as entry metadata.
- REPL `_inline_review` test left as the follow-up you OK'd.

## 🟡 Nits

- `core/undo.py`: `tmp_name = tmp.name` now precedes `tmp.write(payload)`; new test injects a `write` failure and asserts the temp file is unlinked (fails on the old order).
- `not isinstance(revlog_id, bool)` guard added (`"5000"`, `5000.0`, `True`, `None` all covered by test).
- `revlog_id`→`revlog_deleted` contract note is now in the PR description, along with the new `revlog_id` field on `review:answer` output.
- Both hint-bar tests assert the exact `.plain` string (catches key/label swaps, dropped separators, and hidden bindings leaking).
- `test_action_rate_successful_answer_pushes_undo` now pins snapshot-before-answer (fake `answer_card` mutates `Store.queue`; the pushed snapshot shows the pre-answer value); failure test gained `_answered == 0` and a `_load_next` counter.
- Exception text through Rich markup: `escape()` applied at `repl.py` (incl. the two flagged pre-existing sites) and the `review_app.py` status messages I touched.

## 💡 Suggestions

- Taken: documented the semantics in README Safety Notes and SKILL.md.
- Deferred (happy to do in a follow-up): `answer_with_undo` helper, `DirectBackend` pass-throughs, `hint_bar` pure function, fsync — all reasonable but larger refactors than this PR's scope.
- Rationale correction applied to the code comment and the PR description (fix 4 section rewritten; contract note added).

Verified: `pytest` 655 passed, `pytest -m tui` 67 passed, `ruff check anki_cli` clean, `ty check anki_cli` clean.

ubermenchh (approved):
# Re-review of `63e5805`

**Verdict:** APPROVE
**Counts:** 🟠 0 important · 🟡 1 nit · 💡 3 suggestion · ✅ 3 praise

## Summary

All three Important findings are resolved in the way I'd hoped: `answer_card` returns the `revlog_id` it wrote, the three push sites carry it in the snapshot, and `restore_card_state` deletes exactly `WHERE id = ? AND cid = ? AND usn = -1`. The time-window logic and the duplicate timestamp are gone, so the `>`/`>=` boundary and ms-vs-seconds hazards have no code left to attack. The post-commit `push` is best-effort at all three sites, `_save` captures `tmp_name` before writing, and the new tests are real regression guards rather than characterization tests. Thanks for the thorough turnaround and for keeping it as a separate commit — that made the delta easy to verify.

## Verification

- `63e5805`: `pytest` 655 passed, `pytest -m tui` 67 passed, `ruff` and `ty` clean.
- Test-merge with current `main` (`dc822e0`, includes #53 and #54): conflict-free; merged `restore_card_state` keeps #53's `odid`/`odue` restore alongside the new delete; 681 + 67 green.
- Mutation check, 9 mutants against the new code — **8 killed**:

| Mutant | Killed by |
|---|---|
| drop `AND usn = -1` | `test_restore_card_state_does_not_delete_synced_revlog_row` |
| drop `AND cid = ?` | `test_restore_card_state_revlog_id_scoped_to_card` |
| `answer_card` stops returning `revlog_id` | `test_answer_card_updates_card_and_writes_revlog_non_lapse` |
| CLI: push moved back above `answer_card` | `test_review_answer_direct_failure_leaves_no_undo_entry` |
| CLI: remove `try/except OSError` around push | `test_review_answer_direct_undo_push_failure_still_succeeds` |
| TUI: remove `try/except OSError` around push | `test_action_rate_undo_push_failure_still_counts_answer` |
| TUI: snapshot taken after `answer_card` | `test_action_rate_successful_answer_pushes_undo` (stateful fake) |
| `_save`: revert `tmp_name`/`write` order | `test_save_write_failure_removes_tmp_file` |
| drop `not isinstance(revlog_id, bool)` | **survives** — see nit |

- `rich.markup.escape` output is handled correctly by Textual 7.5's `Content.from_markup` (an unescaped `[/x]` in an exception message would previously have raised `MarkupError` inside the except handler).

## 🟡 Nit

**`test_restore_card_state_ignores_non_int_revlog_id` — the `True` case is vacuous** — `tests/integration/test_anki_direct_snapshot_restore.py:563-572`. sqlite binds `True` as `1`, and no fixture row has `id = 1`, so the delete matches nothing whether or not the bool guard exists. `"5000"` and `5000.0` do bite (SQLite integer affinity would match row 5000 without the `isinstance(int)` half), so only the bool half is untested. Fix: `_insert_revlog_row(db_path, row_id=1, cid=100)` before the loop and assert survivors `[1, 5000]`. Fine to do post-merge.

## 💡 Suggestions (follow-ups, none blocking)

- **Surface the "already synced" outcome.** After a sync, undo restores the card row but (correctly) leaves the revlog row, and the TUI/REPL just say "undone". The CLI JSON exposes `revlog_deleted: 0`, but the interactive surfaces could say `undone (review already synced; kept in history)` when the snapshot carried an int `revlog_id` and nothing was deleted. Keeping the row is the right call — deleting a synced row would diverge and the row would resurrect on the next full download.
- **`review.py`: construct `UndoStore()` before `answer_card`.** It's currently built inside the post-commit window and only `OSError` is caught; `Path.expanduser()` raises `RuntimeError` when `HOME` is unset and the uid has no passwd entry, which would escape and fail the command after the review is durable. Very unlikely in practice; moving the constructor next to the snapshot call (or catching `RuntimeError` too) closes it. `repl.py` and `ReviewApp` already construct the store up front.
- **Pre-existing: `snapshot_card_state` sits outside the answer `try` in both TUI paths** (`repl.py:487`, `review_app.py:468`). Direct-backend reads are allowed while Anki Desktop is open, so a card deleted between render and rating raises `LookupError` out of the loop/action and takes the REPL/app down. Moving the snapshot call into the same `try` gives "answer failed: Card not found" instead. Not introduced here.

## ✅ Praise

- **Exact-row delete + `usn = -1` guard** (`anki_direct.py:1162-1168`). The `revlog_id` returned is the very value bound in the INSERT, so the undo target cannot drift; the usn guard also makes stale entries safe after a full-sync download. `test_restore_deletes_the_undone_reviews_revlog_row` deliberately leaves `_allocate_epoch_ms_id` unpatched, which pins that restore no longer depends on snapshot timing.
- **Test design.** The stateful fake store in `test_action_rate_successful_answer_pushes_undo` (answer mutates `queue`) is the right way to pin snapshot-before-answer; the exact-string hint-bar assertions kill the swap/separator/hidden-binding mutants in one line; `test_save_write_failure_removes_tmp_file` fails on the old ordering.
- **Docs and rationale.** README/SKILL.md now state the undo semantics including the synced-row behaviour, and the code comment at `anki_direct.py:1153-1158` now gives the correct cause (the review's own ease 1..4 row surviving).

The deferred refactors (`answer_with_undo` helper, `DirectBackend` pass-throughs, `hint_bar()`, `_inline_review` test) are reasonable follow-ups; I'd suggest opening an issue so they don't get lost.

Issue #25, opened 2026-09-12
preprocess_argv mangles any option value containing '=' (breaks --query "prop:lapses=0")


ubermenchh:
## Problem

`preprocess_argv` (`anki_cli/cli/params.py:27-49`) rewrites **every** non-dash token containing `=` into `--key value`, including tokens that are the *value* of a preceding option. `NamespaceGroup.parse_args` (`cli/app.py:30-32`) applies it to the full argv, so every subcommand is affected.

## Reproduce (verified)

```sh
$ anki --format json cards:ids --query "prop:lapses=0"
Usage: anki cards:ids [OPTIONS]
Error: Got unexpected extra argument (0)
```

This is the exact example from `SKILL.md:219`. The token becomes `--query --prop:lapses 0`.

Also broken:
- `note:add --Back "a=b"` (any field value containing `=`)
- `config:set --value "x=y"`
- `notetype:css --css "div { width=…"` or anything CSS-like
- Deck names containing `=`

## Suggested fix

Only rewrite a token when the **previous** token is not an option that takes a value. Since `NamespaceGroup` knows the resolved `click.Command`, inspect `cmd.params` to know which options consume a value:

```python
value_opts = {o for p in cmd.params if isinstance(p, click.Option) and not p.is_flag for o in p.opts}
out = []
prev = None
for tok in argv:
    if prev in value_opts or tok.startswith("-") or "=" not in tok:
        out.append(tok)
    else:
        key, _, val = tok.partition("=")
        out += [f"--{key}", val]
    prev = tok
```

Alternatively restrict `key=value` sugar to the REPL only.

Add parametrized tests in `tests/unit/test_params.py` for `["--query", "prop:lapses=0"]`, `["--deck", "Lang=Spanish"]`, `"a==b"`, `"Front=--"`, and a bare `"-"`.

PR #50, opened 2026-09-13
fix(direct): fail-closed write guard, Anki-literal csum; cover _ensure_write_safe (#23, #38, #62)

### LINKED ISSUE #38: tests: write guard, FSRS mapping, and rollback are never executed; checksum oracle mirrors impl
## Problem

`anki_cli/db/anki_direct.py` shows 86% line coverage, but the two pieces of code that matter most for a user's collection are **never executed by any test**:

### 1. `_ensure_write_safe` is bypassed in 41 tests
`db/anki_direct.py:2223-2230` is monkeypatched to `lambda: None` in every integration test that writes. Its primitives (`_sqlite_write_locked`, `_anki_process_running`) are unit-tested in `tests/backends/test_detect.py`, but the guard that actually refuses to write while Anki is open has zero coverage. Mutation check: change `or` → `and` at `:2226`, or delete the `raise` — every test stays green.

### 2. The FSRS → Anki mapping is only ever stubbed
`_card_row_to_fsrs` (`:2515-2565`) and `_map_fsrs_result_to_anki` (`:2567-2600`) compute `due`, `ivl`, `left`, `type`, `queue` that get written to the collection. `tests/integration/test_anki_direct_answer_card.py:151-185` and `test_anki_direct_preview_ratings.py:127-163` replace them (and `_build_scheduler`) with stubs, so only the SQL glue is tested.

Mutations that pass today:
- `max(1, round(...))` → `int(...)` at `:2583`
- `- crt_day` → `+ crt_day` at `:2582`
- swap `State.Relearning` / `State.Learning` at `:2549`
- `remaining * 1000 + remaining` → `remaining` at `:2590`

Root cause: `datetime.now(UTC)` at `:2576` is not injectable, so the author stubbed the whole function.

### 3. No atomicity test for `_connect_write`
The `ROLLBACK` branch (`:74-76`) is never reached. `rg "ROLLBACK|IntegrityError|OperationalError" tests/integration` → nothing.

### 4. Checksum oracle mirrors the implementation
`tests/integration/test_anki_direct_note_mutators.py:13-15` re-implements `_field_checksum` line-for-line. It cannot detect that the real Anki checksum strips HTML first.

### 5. `core/search.py` has no direct unit tests
Only ~18 queries exercise it via `test_anki_direct_find_queries.py`. Untested: `_escape_like` (`:664-668`), unterminated quote (`:306`), `flag:8` / `flag:-1` (`:352`), `is:bogus` (`:341`), `*` inside a quoted deck name, trailing backslash.

### 6. AnkiConnect ops tests use invented response shapes
`tests/backends/test_ankiconnect_operations.py:30-35` stubs `_invoke` and returns whatever the test wants. A real `cardsInfo` row has `fields: {name: {value, order}}`, `modelName`, `interval`, etc.; if the parser assumes a different key, stub and code agree and the test passes.

## Suggested fix

1. Two integration tests through a real mutator, no bypass: `_anki_process_running` patched to `True` → `RuntimeError`; and a second connection holding `BEGIN IMMEDIATE` → `RuntimeError`.
2. Inject `now_dt` into `_map_fsrs_result_to_anki` (the sibling already takes one). Table-driven tests with a fixed clock for Review / Relearning / Learning, `col_crt_sec` not a multiple of 86400, and a `next_card.due` in the past.
3. Rollback test: monkeypatch `_field_checksum` to raise after the first `UPDATE`; assert the note row is byte-identical afterward.
4. Replace the checksum oracle with literal values captured from a real collection for `"Q"`, `"<b>Q</b>"`, ``.
5. Unit tests on `tokenize` / `parse` / `compile_card_query`, plus one Hypothesis property: for any printable `q`, `parse(q)` returns a node or raises `SearchParseError` (never `IndexError`), and `sql.count("?") == len(params)`.
6. `tests/backends/fixtures/ankiconnect/*.json` recorded from a real AnkiConnect 6, driven through `httpx.MockTransport`.
7. Run `mutmut` scoped to `anki_direct.py` and `search.py` to rank the rest.

TUI coverage is 10–16% (`repl.py` 10%, `review_app.py` 12%, `browse_app.py` 16%); see #36 and #35 for what to extract so it becomes testable.

### LINKED ISSUE #62: direct: write-guard lock probe fails open on non-lock OperationalError
`_sqlite_write_locked` (anki_cli/backends/detect.py) probes the collection with `BEGIN IMMEDIATE` on a `file:{path}?mode=rw` URI and treats only "locked"/"busy" `OperationalError`s as a held lock — every other `OperationalError` returns `False`, i.e. "not locked, safe to write".

Two consequences:

1. **Fail-open on exactly the collections the guard exists to protect.** A `?`, `#` or `%` in the collection filename corrupts the URI (the query/fragment cuts the path before SQLite sees it), the connect fails with "unable to open database file", and the probe returns `False` — so a direct write proceeds on a collection whose lock state was never checked, while Anki Desktop may hold it.
2. **Unopenable collections are misread as writable.** A directory, a non-SQLite file, or an unreadable path at `collection_path` likewise yields "unable to open" → `False`.

An inconclusive probe should fail closed: the lock state is unknown, not "safe". PR #50 makes `_sqlite_write_locked` re-raise non-lock `sqlite3.Error` and translates it at both callers — `DetectionError` (exit 7, BACKEND_UNAVAILABLE) inside `detect_backend`, `DirectWriteBlockedError` inside `_ensure_write_safe`.

LouisDeconinck:
Closes #38 · Closes #62 · Refs #23 (the csum half; `allow_duplicate` enforcement and stripped `sfld` landed separately in #61 / stay with #23)

## fix(direct): strip HTML before hashing first field for csum (#23 part 1)

`_field_checksum` now runs a port of rslib's `strip_html_preserving_media_filenames` over the first field before hashing: media tags (img/audio/video/object/source) contribute their src/data filename, remaining markup is dropped, and entities are decoded. CLI-written notes now checksum identically to Anki-written ones, so the csum duplicate probe sees `<b>Q</b>` as the same note as `Q`.

Caveat: notes written by earlier anki-cli builds with markup in the first field keep their raw-markup csum until Tools ▸ Check Database rehashes them; duplicate detection may miss them until then (documented in README).

- `_CHECKSUM_MEDIA_TAG_RE`: disjoint alternatives `(?:[^>"']|"[^"]*"|'[^']*')` fix exponential backtracking — a src-less media tag with ~18 quoted attrs previously hung ≥5 s inside the write lock; n=200 now takes ~0.06 ms. Accepted divergence: a quote character that is not part of a balanced pair inside a media tag.
- `_CHECKSUM_HTML_TAG_RE`: `<style[^>]*>`/`<script[^>]*>` are match-equivalent to the old `.*?` forms (first `>` wins) but single-parse — kills the cubic unclosed-`<style>` case.
- Entities: `html.unescape` was lenient (partial decode + HTML5 no-semicolon legacy forms). `_decode_entities_strict` mirrors htmlescape `decode_html`'s all-or-nothing contract: any malformed entity (bare `&`, unknown name, bad numeric, surrogate, uppercase `&#X`) keeps the ENTIRE original string. Named lookup uses the HTML4 `name2codepoint` table, verified identical to htmlescape's `NAMED_ENTITIES` (252 names, same codepoints).
- Skips the regex passes when the field has no `<` (rslib borrowed-Cow fast path), reuses the already-computed csum in `add_note`, and makes `_field_checksum` a staticmethod (it uses no self).
- Tests: pinned to rslib's own vectors (`"test"` → 2840236005, `"今日"` → 1464653051) plus literal-oracle vectors covering every stripper branch (media tag names, quote styles, comments, style/script bodies, decode and keep-as-is entity paths, mutation-survivor rows for the regex details), a linear-time regression test for the backtracking bug, and the cross-markup csum-duplicate refusal. Shared `CSUM_*` oracle constants live in `tests/integration/conftest.py`.

## fix(direct): fail closed in the write guard's lock probe (#38, #62)

- `_ensure_write_safe` raises `DirectWriteBlockedError` (a `RuntimeError` subclass) so the refusal is catchable by type instead of message text.
- `_sqlite_write_locked` built its probe URI with `f"file:{path}?mode=rw"`: a `?`, `#` or `%` in the collection path corrupted the URI, the connect failed "unable to open database file", and the probe returned `False` — the lock half of the guard silently passed on a collection it never probed (filed as #62). Now uses `Path.as_uri()` (percent-encodes) and re-raises unexpected `sqlite3.Error`s: an inconclusive probe must fail closed, not open.
- The fail-closed re-raise is translated at both boundaries, catching `sqlite3.Error` (not just `OperationalError`, so a non-SQLite file — "file is not a database", a `DatabaseError` — is covered too):
  - `detect.py` `_probe_write_lock`, used by both `detect_backend` call sites → `DetectionError(exit_code=7)` → `BACKEND_UNAVAILABLE` envelope;
  - `anki_direct.py` `_ensure_write_safe` → the typed `DirectWriteBlockedError` refusal.

**Behavior change:** an unopenable `--col` target (e.g. a directory or a non-SQLite file) now fails detection with exit 7 / `BACKEND_UNAVAILABLE`. On main the probe swallowed "unable to open" as "not locked" and the command proceeded; the first revision of this PR re-raised it as a raw `sqlite3` traceback out of the Click group for every subcommand.

## test(direct): cover `_ensure_write_safe` refusal paths (#38)

The guard was monkeypatched away in every write-path test, so the `or`/`raise` had no coverage. These keep it intact:

- refuse via `_anki_process_running → True` and via a second connection holding `BEGIN IMMEDIATE` (real `_sqlite_write_locked` probe, not a stub), plus a spy-checked pass-through proving both primitives are consulted — the order pin is deliberate (cheap process check before `BEGIN IMMEDIATE` on the live collection);
- mid-flight failures (checksum error; tags failure after the fields UPDATE) leave the note byte-identical and `col.mod` untouched — the atomicity contract, regardless of whether the revert is the except-path `ROLLBACK` or `conn.close()` discarding the open transaction;
- probe-inconclusive refusals: a directory as `--col` (`exists()` is True but `sqlite3.connect` cannot open it) raises `DetectionError` exit 7 on the forced and auto detect paths and `DirectWriteBlockedError` on the write path, and the CliRunner end-to-end emits `BACKEND_UNAVAILABLE` with no `Traceback`.

The module docstring records the coupling that makes the `detect_mod` stubs work (`_ensure_write_safe`'s call-time import); `_make_store` is marked TODO(#37).

## Verification

- `ruff check anki_cli tests` clean; `ty check anki_cli` clean
- `pytest`: 1711 passed, 68 tui deselected · `pytest -m tui`: 68 passed (1779 total)
- Round-2 repro now emits the envelope: `anki --backend direct --col <dir> version` → exit 7, `"BACKEND_UNAVAILABLE"`, no `Traceback` (pinned by `test_cli_unopenable_collection_emits_backend_unavailable`)
- Mutation checks on `_ensure_write_safe` and the stripper (round 1): `or` → `and` → 3 tests red; `raise` removed → refusal tests red; dropping the HTML strip → csum tests red


ubermenchh (changes requested):
## Review — write-guard coverage + csum HTML-stripping

**Verdict:** Request changes
**Counts:** 🔴 0 blocking · 🟠 5 important · 🟡 7 nit · 💡 6 suggestion · ✅ 3 praise

### Summary

The test work is good and genuinely closes #38: the write guard and the transaction ROLLBACK are exercised for real, and the tautological sha1 oracle is gone. But the PR also carries a production behavior change under a `test:` commit type — `_field_checksum` now runs a regex port of rslib's HTML stripper on every note write — and that port has a measured exponential-backtracking bug: a ~100-byte `` tag with ~18 quoted attributes and no `src=` hangs the process for 5 s+ while holding the collection write lock (Rust's regex engine is linear; Python's isn't). The fix is a verified one-line regex change. Beyond that, the entity-decode step diverges from rslib on inputs mixing valid and bare `&`, and the stripper's other branches have no tests. Land the regex fix, retype the commit, add the missing vectors (rslib ships two independent ones the PR already passes), and this is a solid merge.

All timings and semantic claims below were verified by running the PR branch or reading the upstream rslib / htmlescape source.

---

### 🟠 Important

#### 1. Fix exponential backtracking in `_CHECKSUM_MEDIA_TAG_RE`
**`anki_cli/db/anki_direct.py:38-50`** — a media tag with many quoted attributes and no `src=`/`data=` makes the regex explore 2ⁿ parses before failing.

The group `(?:[^>]|"[^"]+?"|'[^']+?')+?` is ambiguous: each `"a"` can be consumed as three `[^>]` chars or one quoted token. When `\b(?:src|data)\b=` never matches, the engine enumerates every split. Measured on this branch with `"10 s (killed) |

A realistic 8-attribute src-less `` is 0.55 ms, so ordinary notes are unaffected; the trigger is a heavily-attributed `<video controls="" autoplay="" … data-f="6"><source src="clip.mp4">` from pasted web HTML or a shared deck. Reachable via `add_note` (which hashes twice, L1873 and L1908), `update_note`, `add_notes` bulk import (the per-item `except Exception` doesn't help — nothing raises), and `remove_notetype_field`, which rehashes every note of the notetype inside one `BEGIN IMMEDIATE`. Ctrl-C recovers cleanly (`finally: conn.close()` discards the transaction), so no data loss — which is why this is 🟠 rather than 🔴. It still needs to land before merge.

Secondary, same class, lower plausibility: `_CHECKSUM_HTML_TAG_RE`'s `<style.*?>.*?</style>` is cubic on repeated *unclosed* `<style>`/`<script>` tags — measured `"<style>"*400` = 500 ms, `*1600` = >8 s. Needs kilobytes of malformed input, but the fix is free.

**Suggested fix** (verified: n=200 → 0.06 ms; byte-identical output on rslib's own test vectors ``, `<html>`, `<html>`, plus `alt="a>b"`, `data=z.swf`, unquoted `src=y.png`, mixed quotes; only divergence is an *unbalanced* quote inside a media tag, e.g. `<img alt="a src=x.png>` → rslib ` x.png `, fix `` — worth a comment as accepted):

```python
_CHECKSUM_MEDIA_TAG_RE = re.compile(
    r"""
    <\b(?:img|audio|video|object|source)\b
    (?:[^>"']|"[^"]*"|'[^']*')+?      # disjoint alternatives: a quote char is only consumable by the quoted branch
    \b(?:src|data)\b=
    (?:
        "([^"]+?)"[^>]*>
        |'([^']+?)'[^>]*>
        |([^ >]+?)(?:\x20[^>]*>|>)
    )
    """,
    re.VERBOSE | re.IGNORECASE | re.DOTALL,
)

# <style[^>]*> is match-equivalent to <style.*?> (first '>' wins either way) but single-parse;
# verified identical on 6 vectors, "<style>"*6400 unclosed drops from >8 s to ~1 s.
_CHECKSUM_HTML_TAG_RE = re.compile(
    r"(<!--.*?-->)|(<style[^>]*>.*?</style>)|(<script[^>]*>.*?</script>)|(<.*?>)",
    re.IGNORECASE | re.DOTALL,
)
```

Plus a regression test (see #4) and reuse `csum` at L1908 instead of re-hashing.

#### 2. Match rslib's all-or-nothing entity decoding
**`anki_cli/db/anki_direct.py:73-74`** — `html.unescape` is lenient; rslib's `htmlescape::decode_html` returns `Err` on any malformed entity and then keeps the **entire original string**.

Verified against `rust-htmlescape/src/decode.rs`: the state machine hits `PrematureEnd` on a bare `&` followed by non-entity text, `UnknownEntity` on `&bogus;`, `MalformedNumEscape` on `&#32a;`; rslib's `decode_entities` then returns the input unchanged. Python decodes what it can and also accepts HTML5 legacy no-semicolon forms (`&nbsp`, `&amp`, `&copy`). Counterexamples (this branch's output measured):

| input | PR hashes | Anki hashes |
|---|---|---|
| `R&amp;D & more` | `R&D & more` | `R&amp;D & more` |
| `a&nbsp b` | `a  b` | `a&nbsp b` |
| `&ampx` | `&x` | `&ampx` |

Effect is limited to duplicate detection missing across CLI-written and Anki-written notes — low blast radius, but parity is the entire point of this change.

**Suggested fix:**

```python
# rslib decode_entities: htmlescape::decode_html is all-or-nothing — any malformed
# entity (bare '&', unknown name, bad numeric) returns Err and the ORIGINAL text is kept.
_ENTITY_RE = re.compile(r"&(#[0-9]+|#[xX][0-9a-fA-F]+|[A-Za-z][A-Za-z0-9]*);")

def _decode_entities_strict(text: str) -> str:
    if "&" not in text:
        return text
    if "&" in _ENTITY_RE.sub("", text):      # a '&' not part of a well-formed entity → Err → keep as-is
        return text
    def _one(m: re.Match[str]) -> str:
        body = m.group(1)
        if body[0] == "#":
            cp = int(body[2:], 16) if body[1] in "xX" else int(body[1:])
            return chr(cp)
        return html.entities.html5[body + ";"]
    try:
        return _ENTITY_RE.sub(_one, text).replace("\xa0", " ")
    except (KeyError, ValueError, OverflowError):
        return text
```

Caveat to note in a comment: htmlescape's table is HTML4 (~250 names); `html.entities.html5` has 2231, so HTML5-only names like `&check;` still diverge unless the lookup is restricted. Add `"R&amp;D & more"` and `"Q&amp;A"` to the vector table.

#### 3. Retype the commit and document the csum behavior change
**`anki_cli/db/anki_direct.py:36-76, 2586-2591`** — commit `16cfbd6` is titled `test(direct): …` but changes the value written to `notes.csum` on every note-write path.

Changelog tooling, `git bisect`, and anyone hunting "why did csums change" will skip a `test:` commit; if a dup-detection regression appears, the revert target is mislabelled as tests-only. Separately, notes written by earlier anki-cli builds with markup in the first field keep their raw-markup csum, so they're invisible to the dup probe until Anki's *Check Database* rehashes them — correct behavior, but undocumented.

**Suggested fix:** split into `fix(direct): strip HTML before hashing first field for csum (#23 part 1)` (production change + literal tests) and `test(direct): cover _ensure_write_safe refusal paths (#38)` (guard file), or at minimum retitle to `fix(direct):` and lead the body with the behavior change. Add one line to the README's direct-backend caveats: *"csum for notes with HTML in the first field now matches Anki. Rows written by older anki-cli versions keep their old csum until Tools ▸ Check Database recomputes them; duplicate detection may miss them until then."*

#### 4. Cover the stripper's untested branches with an external oracle
**`anki_cli/db/anki_direct.py:52-56, 71-74`**, **`tests/integration/test_anki_direct_note_mutators.py:374-400`** — three vectors exercise `<b>` and double-quoted `` only.

Mutations that survive today: drop `re.IGNORECASE`; delete the `unescape` branch or the `\xa0` replace; delete any of the `<!--`/`<style>`/`<script>` alternations; break single-quoted or unquoted `src`; break `audio|video|object|source`; break `data=`; and the ReDoS in #1. The literals are labelled "captured from a real Anki collection" but equal the PR's own implementation on every input, so the next maintainer can't re-verify the claim. rslib ships two independent vectors in `rslib/src/notes/mod.rs::test_field_checksum` — `"test"` → `2840236005`, `"今日"` → `1464653051` — and this branch passes both; those belong in the table as the external oracle.

```python
@pytest.mark.parametrize(("first_field", "expected"), [
    ("test", 2_840_236_005),                    # rslib vector (rslib/src/notes/mod.rs)
    ("今日", 1_464_653_051),                     # rslib vector
    ("Q", _CSUM_Q), ("<b>Q</b>", _CSUM_Q), ('<img src="x.png">', _CSUM_IMG_X_PNG),
    ("<IMG SRC='y.jpg'>", ...), ("<img src=z.gif>", ...),     # case-fold, single-quoted, unquoted
    ('<audio src="a.mp3"></audio>', ...), ('<object data="o.swf">', ...),
    ("<!-- c -->Q", _CSUM_Q), ("<style>p{}</style>Q", _CSUM_Q), ("<script>x()</script>Q", _CSUM_Q),
    ("Q&amp;A", ...), ("a&nbsp;b", ...), ("R&amp;D & more", ...),   # entity paths incl. strict-decode case
])
def test_field_checksum_matches_anki(first_field, expected, tmp_path): ...

def test_field_checksum_srcless_media_tag_is_linear(tmp_path):
    """Regression for the backtracking bug: 40 quoted attrs, no src, must strip to '' fast."""
    attrs = " ".join(f'a{i}="v{i}"' for i in range(40))
    store = AnkiDirectReadStore(_touched(tmp_path))
    t0 = time.perf_counter()
    assert store._field_checksum(f"<img {attrs}>Q") == _CSUM_Q
    assert time.perf_counter() - t0 < 0.5
```

#### 5. Add the dup-detection test this change is for
**`anki_cli/db/anki_direct.py:1873-1886`** — the only user-visible effect of stripping (`<b>Q</b>` now collides with `Q`) has no test; `_insert_note(csum=0)` removed the last fixture that could have exposed it.

No test anywhere exercises the `dup_ids and not allow_duplicate` branch (both `add_note` callers write into an empty table). Deleting the SELECT, inverting `not allow_duplicate`, or flipping `csum = ?` to `!=` all pass today.

```python
def test_add_note_warns_on_csum_duplicate_across_markup(monkeypatch, tmp_path, capsys):
    store, db_path = _make_store(tmp_path)
    monkeypatch.setattr(store, "_ensure_write_safe", lambda: None)
    _insert_note(db_path, note_id=1, front="Q", back="A", csum=_CSUM_Q)
    store.add_note(deck="Default", notetype="Basic",
                   fields={"Front": "<b>Q</b>", "Back": "A"}, tags=None, allow_duplicate=False)
    assert "duplicate note detected" in capsys.readouterr().err
    store.add_note(deck="Default", notetype="Basic",
                   fields={"Front": "<b>Q</b>", "Back": "A"}, tags=None, allow_duplicate=True)
    assert capsys.readouterr().err == ""
```

---

### 🟡 Nits

- **`tests/integration/test_anki_direct_note_mutators.py:419`** — `assert note["sfld"] == "<b>Q</b>"  # sfld keeps raw field text` pins behavior that is wrong vs Anki: rslib `prepare_for_update` stores `strip_html_preserving_media_filenames(sort_field)`, so Anki writes `sfld="Q"`. Pre-existing and explicitly deferred to #23, but don't cement it — `# TODO(#23): Anki stores stripped sfld ("Q"); flip when fixed`.
- **`tests/integration/test_anki_direct_write_guard.py:194-234`** — docstrings say "must reach its ROLLBACK branch"; I deleted `anki_direct.py:115-116` and all 5 tests still pass, because `conn.close()` in `finally` discards the open transaction. The tests prove the right contract (no partial write persists, `col.mod` untouched) — say that instead.
- **`tests/integration/test_anki_direct_write_guard.py:184`** — `consulted == ["running", "locked"]` fails if the `or` operands are reordered, which changes no behavior. Either `sorted(consulted) == [...]`, or add a docstring line stating the order is deliberate (cheap process check before probing the live collection with `BEGIN IMMEDIATE`).
- **`anki_cli/db/anki_direct.py:1908`** — `csum` is already computed at L1873; pass it here instead of calling `_field_checksum` again.
- **`tests/integration/test_anki_direct_write_guard.py:6-8, 130`** — all `monkeypatch.setattr(detect_mod, …)` stubs work only because `_ensure_write_safe` does a call-time `from … import` (`anki_direct.py:2560`). `detect.py` has no `anki_cli` imports, so there's no cycle forcing that; if someone hoists it, the process-running test fails loudly but the lock and ROLLBACK tests become host-dependent. Add a module-header comment naming the coupling, or hoist now and patch `anki_cli.db.anki_direct._anki_process_running`.
- **`tests/integration/test_anki_direct_write_guard.py:30-113`** — this is the 17th file-local `_make_store`. Deferring to #37 is right; add `# TODO(#37)` and drop the `notetypes` DDL (nothing here reads it; `fields` is needed by `_field_schema_for_mid`).
- **`tests/integration/test_anki_direct_notetype_mutators.py:476, 567, 625`** — `2_264_392_759` appears twice with duplicate comments; the oracle set is split across two files with two conventions.

---

### 💡 Suggestions

- **Make the checksum a module function, outside the 3.3k-line store** (`anki_direct.py:38-76, 2586-2591`). `_field_checksum` uses no `self`; tests build a store on a `touch()`ed empty file just to call it. Move lines 38-76 plus `def field_checksum(first_field: str) -> int` into `anki_cli/db/anki_text.py`; replace `self._field_checksum(` at 661/1873/1908/2010. The deferred #23 sfld work needs the same function. Minimum viable: `@staticmethod`.
- **Put the csum oracle constants in one place** (`conftest.py`): `CSUM_Q`, `CSUM_A`, `CSUM_B`, `CSUM_F1`, `CSUM_IMG_X_PNG`, `CSUM_TEST`, `CSUM_KYOU` under *"Anki-written csum values. Never derive these from the CLI's own checksum — they are the oracle."* The bare `2_264_392_759` is one "cleanup" away from becoming `store._field_checksum("a")` again.
- **Raise a named exception instead of matching message text** (`anki_direct.py:2563`, `test_anki_direct_write_guard.py:27`): `class DirectWriteBlockedError(RuntimeError)` lets the CLI layer exit cleanly on "Anki is open" without string matching; tests drop `_GUARD_MSG`.
- **Skip the regex passes when there's no `<`** (`anki_direct.py:71`): `if "<" in text:` mirrors rslib's borrowed-`Cow` fast path and is provably output-neutral.
- **Pre-hash normalization gap (pre-existing, out of diff)** (`anki_direct.py:1873, 2010`): rslib `normalize_field` strips ASCII control chars (except `\n`/`\t`) and NFC-normalizes before computing csum/sfld. `"cafe\u0301"` (NFD) gets a different csum than Anki would produce. Worth a follow-up issue alongside #23.
- **`_sqlite_write_locked` fails open on non-lock errors (pre-existing)** (`detect.py:280-285`): `f"file:{db_path}?mode=rw"` isn't percent-encoded; a path with `?`/`#`/`%` yields "unable to open database file" → returns `False` → lock half of the guard silently passes. Use `db_path.resolve().as_uri() + "?mode=rw"` and re-raise unexpected `OperationalError`s.

---

### ✅ Praise

- **The oracle now points at Anki, not at the implementation** (`test_anki_direct_note_mutators.py:12-17`). Deleting the sha1 copy and asserting on-disk literals is what turned a tautology into a test, and it surfaced the markup-vs-text divergence. The port passes rslib's own two upstream vectors, which is the strongest evidence in this PR that the algorithm is right.
- **The two-statement ROLLBACK test proves atomicity, not just "nothing ran"** (`test_anki_direct_write_guard.py:219-234`). Letting the `flds/sfld/csum` UPDATE execute before `_format_tags` raises is the case that distinguishes a real transactional revert from an early exit; it's the only test here that would catch `COMMIT` drifting into `finally`.
- **Real second-connection lock instead of a stubbed probe** (`test_anki_direct_write_guard.py:134-157`). Deterministic (RESERVED lock vs 50 ms busy timeout; measured 0.06 s), leak-safe via `finally`, and cross-checked by the pass-through test — mutate the probe to constant `True` and test 3 goes red, constant `False` and this one does.

---

### Follow-ups worth tooling

- **CodeQL `py/redos` / `py/polynomial-redos`** on `anki_direct.py:38-56` after the fix, and on `core/template.py:5-7` (`_SECTION_RE`/`_CLOZE_RE` use DOTALL `.*?` with a backreference on the same untrusted content at render time; not audited here).
- **Hypothesis parity fuzz** on `_strip_html_preserving_media_filenames` against the `anki` pip package's `anki.utils.field_checksum` for a pinned version — the honest replacement for the "verified against rslib's test vectors" claim, which shipped none of those vectors.
- **Scoped `mutmut`** on `anki_direct.py:59-76,2586-2591` + the new test file, to replace the manual "3 tests red" claim with numbers (it missed that the ROLLBACK statement itself is unpinned).

*Limits of this review: htmlescape semantics were read from source, not executed against rslib; the HTML5-vs-HTML4 entity-table point in fix #2 is reasoned, not tested.*

LouisDeconinck:

LouisDeconinck:

LouisDeconinck:
Thanks for the thorough review — all items addressed in the reworked branch (now 3 commits: `55a3e06` fix csum, `7270909` guard hardening, `febf4c1` guard tests).

**🟠1 ReDoS** — `_CHECKSUM_MEDIA_TAG_RE` replaced with your disjoint-alternatives regex verbatim (`(?:[^>"']|"[^"]*"|'[^']*')+?`), including the accepted unbalanced-quote divergence as a comment. `_CHECKSUM_HTML_TAG_RE` now uses `<style[^>]*>`/`<script[^>]*>` with the match-equivalence note. Regression test `test_field_checksum_srcless_media_tag_is_linear` added (40 quoted attrs, no `src`, <0.5s — measured 0.04ms), and `add_note` reuses the already-computed `csum` instead of re-hashing.

**🟠2 Entity decode** — implemented the all-or-nothing `_decode_entities_strict`, but with three tightenings verified against htmlescape's actual `decode.rs`/`entities.rs` source rather than the sketched version:
- Named lookup uses `html.entities.name2codepoint` — I diffed it against htmlescape's `NAMED_ENTITIES` and they are **identical** (252 names, same codepoints), so the `&check;` HTML5-only divergence you flagged is closed, not just documented.
- `#x` is lowercase-only upstream: `&#X41;` → MalformedNumEscape in htmlescape (`Numeric if c == 'x'`), so the regex is `#x[0-9a-fA-F]+`, not `#[xX]`.
- `&#xD800;`-style surrogates err via `char::from_u32` → InvalidCharacter; Python's `chr()` accepts them and would crash the UTF-8 encode in the sha1 — explicitly rejected.

All the counterexample rows now hash the kept-original string: `R&amp;D & more`, `a&nbsp b`, `&ampx`, plus added `&check;`/`&#X41;`/`&#xD800;`/`&bogus;`/`&#32a;`/`&#65;`/`&#x41;` vectors.

**🟠3 Commit retype** — split as suggested: `fix(direct): strip HTML before hashing first field for csum (#23 part 1)` carries the production change + literal tests (body leads with the behavior change), `test(direct): cover _ensure_write_safe refusal paths (#38)` carries the guard file, and the guard-hardening bits went to their own `fix(direct): fail closed in the write guard's lock probe (#38)`. README Safety Notes got the old-csum caveat line verbatim.

**🟠4 Vectors** — rslib's `"test"` → 2840236005 and `"今日"` → 1464653051 are in the table as the external oracle, plus literals for every stripper branch (all 5 media tag names, case-folded, single/double/unquoted `src`/`data`, comment/style/script bodies, decode + keep-as-is entity paths).

**🟠5 Dup detection** — `test_add_note_warns_on_csum_duplicate_across_markup` added verbatim.

**🟡 nits** — sfld assert now carries `# TODO(#23)`; ROLLBACK docstrings reworded to pin the contract (no partial write persists, `col.mod` untouched) rather than the mechanism; the `consulted` order pin is documented as deliberate; module docstring records the call-time-import coupling for the `detect_mod` stubs; `_make_store` marked TODO(#37). One correction: `notetypes` DDL is **not** dead — `update_note` → `_field_schema_for_mid` selects `notetypes.config` for the sort index, so it stays (comment added).

**💡 taken** — `DirectWriteBlockedError(RuntimeError)` raised by the guard (tests dropped `_GUARD_MSG`); CSUM_* oracle constants centralized in `conftest.py`; `_field_checksum` is now `@staticmethod`; `if "<" in text` fast path mirrors the borrowed-Cow skip; `_sqlite_write_locked` now uses `as_uri()` + re-raises unexpected `OperationalError`s (fail closed) — the previously-pinned fail-open test was updated and a `col?100%.db` locked-file regression test added. Skipped: the `anki_text.py` move (staticmethod = your stated minimum viable) and the NFC/control-char pre-hash normalization — worth a follow-up alongside #23 as you noted.

`pytest` 678 passed + `-m tui` 57 passed, `ruff check anki_cli` and `ty check anki_cli` clean.

ubermenchh (changes requested):
## Review, round 2 — head `55a3e06` (3 commits on `1cbaf6d`)

**Verdict:** Request changes — narrowly. By count this would be an approve (2 🟠, no 🔴), but #1 is a user-visible regression this revision *introduced*, and it's a 10-line fix. Land #1, refresh the PR body, and this merges.
**Counts:** 🔴 0 · 🟠 2 · 🟡 4 · 💡 3 · ✅ 4

### Round-1 scorecard

Every round-1 item is closed, and the closures were verified on the new head rather than taken from the commit messages:

| Round-1 finding | Status | Verified how |
|---|---|---|
| 🟠 Exponential backtracking in `_CHECKSUM_MEDIA_TAG_RE` | ✅ Fixed | 2000 quoted attrs → 0.63 ms (was >10 s at 20). `<style>*6400` unclosed 0.96 s (was >8 s at 1600) |
| 🟠 Lenient entity decode vs rslib | ✅ Fixed | `_decode_entities_strict` matches htmlescape's `decode.rs` state machine on ~30 adversarial inputs; `name2codepoint` == `NAMED_ENTITIES` (252/252 names and codepoints, compared programmatically) |
| 🟠 `test:` commit hiding a behavior change | ✅ Fixed | Split into `fix`/`fix`/`test`; README caveat added |
| 🟠 Untested stripper branches | ✅ Fixed | 24 vectors incl. rslib's own `"test"`/`"今日"`; all 20 `CSUM_*` literals equal sha1 of the text named in their comment |
| 🟠 No dup-detection test | ✅ Fixed | `test_add_note_warns_on_csum_duplicate_across_markup` |
| 🟡 ×7, 💡 ×6 | ✅ All adopted | staticmethod, csum reuse, `DirectWriteBlockedError`, `<` fast path, TODO(#23)/(#37), docstrings, conftest constants, import-coupling note, `as_uri()` |

Full suite on the head: 678 passed, 4 skipped; ruff clean.

---

### 🟠 Important

#### 1. The fail-closed probe now crashes bootstrap for every command
**`anki_cli/backends/detect.py:290-293`** (commit `7270909`), reached from **`detect.py:59, :83`** and unguarded at **`anki_cli/cli/app.py:138`** — `_sqlite_write_locked` re-raises non-lock `OperationalError`, but two of its three callers are inside `detect_backend()`, which the Click group runs for *every* subcommand, and `app.py` catches only `DetectionError`.

Verified repro:
```
$ anki --backend direct --col /some/dir/collection.anki2 version     # path is a directory
main:  prints the version table
PR:    sqlite3.OperationalError: unable to open database file   (traceback, no JSON envelope)
```
Same on the auto path with a config `collection_path` pointing at an unopenable file. Fail-closed is the right intent for the *write guard*; the bug is that the exception type crosses `detect_backend`, whose only documented failure type is `DetectionError`. Structurally, this slipped because every `detect_backend` test goes through `_patch_detect_helpers` (`tests/backends/test_detect.py:29`), which stubs `_sqlite_write_locked` to `lambda path: locked` — the seam can only ever return a bool in tests, so the new unit test `..._fails_closed` codified a re-raise whose integration consequence was unobservable. (Read-only collections, chmod 444/555, are *not* affected — SQLite falls back and `BEGIN IMMEDIATE` still succeeds; I checked.) `DirectWriteBlockedError` also has zero consumers under `anki_cli/cli`, so the write path still tracebacks on refusal — pre-existing, but the new typed exception was meant to fix exactly that.

**Suggested fix** — translate at both boundaries, catch `sqlite3.Error` (not just `OperationalError`) so a non-SQLite file at the path ("file is not a database", a `DatabaseError`) is covered too:
```python
# detect.py — one choke point used at :59 and :83
def _probe_write_lock(path: Path) -> bool:
    try:
        return _sqlite_write_locked(path)
    except sqlite3.Error as exc:
        raise DetectionError(
            f"Cannot probe the collection lock state at {path}: {exc}", exit_code=7
        ) from exc

# anki_direct.py:_ensure_write_safe
try:
    locked = _sqlite_write_locked(self.db_path)
except sqlite3.Error as exc:
    raise DirectWriteBlockedError(
        f"Cannot verify the collection lock state at {self.db_path}: {exc}"
    ) from exc
if _anki_process_running() or locked:
    raise DirectWriteBlockedError("Anki Desktop appears to be running ...")
```
Plus the two tests that would have gone red:
```python
def test_forced_direct_unopenable_collection_raises_detection_error(tmp_path, monkeypatch):
    bad = tmp_path / "collection.anki2"; bad.mkdir()   # exists() is True; sqlite3 cannot open a directory
    monkeypatch.setattr(detect_mod, "_anki_process_running", lambda: False)
    with pytest.raises(DetectionError) as exc_info:
        detect_backend(forced_backend="direct", col_override=bad)
    assert exc_info.value.exit_code == 7

def test_cli_unopenable_collection_emits_backend_unavailable(tmp_path):
    bad = tmp_path / "collection.anki2"; bad.mkdir()
    result = CliRunner().invoke(cli, ["--backend", "direct", "--col", str(bad), "version"])
    assert result.exit_code == 7
    assert '"BACKEND_UNAVAILABLE"' in result.output and "Traceback" not in result.output
```
Also: `7270909` references `(#38)`, which is a test-coverage issue; the fail-open bug deserves its own issue number so a bisecting maintainer lands on the right context.

#### 2. Refresh the PR title and body to match what will be squash-merged
**PR title/body** — unchanged since round 1. The title is still `test(direct): …` while two of three commits are `fix(direct):` and the diff is now majority production code. The body says 630 tests (now 678), "entities decoded" (now a strict all-or-nothing decoder), and omits the `detect.py` behavior change, `DirectWriteBlockedError`, the ReDoS fix, the README caveat, and the conftest constants.

This is 🟠 rather than 🟡 because `main` squash-merges (all 6 recent commits carry a `(#N)` suffix): the stale body becomes the *only* permanent record, and the three exemplary commit bodies will be discarded on merge.

**Suggested fix:** Title `fix(direct): fail-closed write guard, Anki-literal csum; cover _ensure_write_safe (#23, #38)`. Body: three sections lifting the three commit bodies near-verbatim (they're already the right content), counts updated to 678/4, and an explicit "Behavior change" line for the probe.

---

### 🟡 Nits

- **`tests/backends/test_detect_helpers.py`** (`..._uri_special_chars`) — `col?100%.db` cannot be created on NTFS (`?` is reserved), and `#` — the other URI-truncating character — isn't exercised. CI is ubuntu-only so it passes today; it will be the first test deleted when a Windows leg appears. Parametrize over `["col%100.db", "col#1.db", "col 100.db"]` and keep `?` under `skipif(sys.platform == "win32")` since it's the character that originally broke.
- **`tests/integration/test_anki_direct_note_mutators.py`** (`test_field_checksum_matches_anki`) — mutation survivors against the 24-row table: dropping `re.DOTALL` (no vector has `\n` inside a tag or comment), narrowing `<script[^>]*>` to `<script>` (no attributes on style/script), dropping `video`/`source` from the alternation (only img/audio/object appear). Four rows close them: `'<style type="text/css">\np{}\n</style>Q'`, `'<!-- multi\nline -->Q'`, `''`, `'<video src="v.mp4"></video>'`.
- **`test_add_note_warns_on_csum_duplicate_across_markup`** — pins the warning but not the "warn-only, still inserts" contract; a mutation that `return`s before the INSERT passes the first call. Add `assert _count_notes(db_path) == 2` after the first `add_note`.
- **Branch base `1cbaf6d`** — one behind `main` (#53, filtered-deck fixes). No file overlap; trivial rebase, but re-run the suite after so the count in the body is true for the merge commit.

---

### 💡 Suggestions

- **Divergence shape is broader than the comment says** (`anki_direct.py` `_CHECKSUM_MEDIA_TAG_RE` comment). The comment names `<img alt="a src=x.png>` (unbalanced quote). The more plausible instance is an *unquoted* attribute value containing an apostrophe: `` → rslib ` a.png `, PR `` (verified against rslib's regex). The quoted form `alt="Bob's"` is fine, and Anki's editor always quotes, so this needs hand-written HTML — accepting it is reasonable. Widen the comment to say "any quote character not part of a balanced pair"; if it ever matters, a 3-state NFA scan (outside / in-`"` / in-`'`) reproduces rslib's existence semantics in linear time without a regex.
- **`pytest-timeout` idiom** for `test_field_checksum_srcless_media_tag_is_linear` — the 0.5 s wall-clock bound against 0.6 ms actual is safe; `@pytest.mark.timeout(2)` with 200 attrs is the conventional form if the repo adopts the plugin. Not worth a change alone.
- **Lock probe runs for read-only commands** (`detect.py:59, :83`) — pre-existing design, and it's why #1 surfaces on `version`. With #1's wrapper the cost is bounded to a clean JSON error; leave the redesign for a separate change.

---

### ✅ Praise

- **Commit split and bodies are exactly what a bisector needs** (`55a3e06`, `7270909`, `febf4c1`). `55a3e06` in particular carries the measured ReDoS numbers, the `<style[^>]*>` equivalence argument, the accepted divergence, the migration caveat, and the README line in one commit — `git log -1` on that hash answers every question a maintainer would ask.
- **Strict-decode vector table is one row per htmlescape failure class** (`conftest.py` `CSUM_*_KEPT`, `test_field_checksum_matches_anki` rows 17–24). Bare `&`, missing `;`, HTML5-only name, uppercase `X`, surrogate, unknown name, junk digit — each row isolates one rejection path and hashes the *raw markup*, which is the honest oracle for "decode refused, whole string kept". A leniency regression flips exactly one row and names the branch.
- **Entity decoder parity holds under adversarial input.** Beyond the table, I ran `&#;`, `&;`, `&#x;`, `&Amp;`, `a & b;`, `&#0;`, `&#1114112;`, `&#4294967296;`, `&#99999999999999999999;`, `&frac12;`, `&#X41;`, `&amp;&bogus;`, `&lt;b&gt;`, `x\xa0y&amp;`; the specialist added `&am p;`, `&a#b;`, `&#-1;`, `&#+1;`, `&#1x;`, `&&amp;`, `&#38;#38;` (no double-decode), `&#x0000000000041;`, `&#xFFFFFFFFF;`. Zero divergences from the `decode.rs` state machine. The `_ENTITY_RE.sub("", text)`-then-check-for-`&` trick is a compact and correct stand-in for the all-or-nothing contract.
- **URI special-chars test proves the old code failed for the right reason.** Under main's `f"file:{path}?mode=rw"`, SQLite parses the query from the first `?`, truncates the path, and the probe returns `False` on a locked file — so `is True` goes red on old code and green on new. Filename portability aside, the shape is exactly right.

---

### Not re-run this round

security / performance / architecture specialists — the ReDoS is measured-fixed across four input shapes, and every round-1 architecture suggestion was adopted. Re-running them on this delta would be padding.

*Limits: rslib parity claims are against `ankitects/anki@main` `rslib/src/text.rs` and `notes/mod.rs`, and `veddan/rust-htmlescape@master` `decode.rs`/`entities.rs`, read from source in this session; I did not execute rslib.*

LouisDeconinck:
Round-2 findings addressed at `af5e06a` (was `55a3e06`).

**🟠 #1 — fail-closed probe crashing bootstrap: fixed.**

- `detect.py`: new `_probe_write_lock` choke point used by both `detect_backend` call sites (:59, :83). It catches `sqlite3.Error` — not just `OperationalError`, so "file is not a database" is covered — and raises `DetectionError(exit_code=7)`, which `app.py:138` translates to `BACKEND_UNAVAILABLE`.
- `anki_direct.py` `_ensure_write_safe`: probe errors raise `DirectWriteBlockedError("Cannot verify the collection lock state at …")`; the `_anki_process_running() or locked` refusal is unchanged.
- Tests (the ones that would have gone red): `test_forced_direct_unopenable_collection_raises_detection_error` plus an auto-path twin for a config `collection_path`, `test_cli_unopenable_collection_emits_backend_unavailable` (CliRunner: exit 7, `"BACKEND_UNAVAILABLE"`, no `Traceback`), and `test_write_refused_when_lock_probe_is_inconclusive` pinning the write-path half of the translation.
- Filed #62 for the pre-existing fail-open bug so a bisect lands on the right context; `f8c0be9` references it.

**🟠 #2 — title/body refreshed.** Three sections lift the three commit bodies near-verbatim, explicit "Behavior change" line for the probe, counts updated.

**Merge with main** (was CONFLICTING after #49/#53–#56/#59–#61): union imports in `anki_direct.py`, both README Safety-Notes bullets, both test suites in `test_anki_direct_note_mutators.py` (the csum-duplicate test adapted to #61's raise-not-warn semantics, which is the stronger pin). Two pieces of merge fallout repaired in the last two commits: the #25 argv-sugar regression test was silently dropped by the conflict resolution (restored verbatim), and the write-guard `_make_store` needed `decks`/`templates`/`cards`/`graves` plus the `generated_cards` key now that #59's `update_note` runs `_generate_missing_cards`. Also adopted the URI-special-chars parametrization (percent/hash/space, `?` under `skipif(win32)`).

Suite on the head: `pytest` 1711 passed + `pytest -m tui` 68 passed (1779 total); `ruff check anki_cli tests` and `ty check anki_cli` clean.

umangk-2410 (approved):
## Review, round 3 — head `af5e06a` (merge-base = current `main` `f3a0bfa`)

**Verdict:** Approve
**Counts:** 🔴 0 · 🟠 0 · 🟡 2 · 💡 1 · ✅ 3

### Round-2 scorecard

Both 🟠 findings and all four 🟡 are closed; each verified on the head rather than from the comment:

| Round-2 finding | Status | Verified how |
|---|---|---|
| 🟠 #1 fail-closed probe crashing bootstrap | ✅ Fixed | `--col <directory> version` → `BACKEND_UNAVAILABLE`, exit 7, no traceback. Also checked the `DatabaseError` path the wrapper's `sqlite3.Error` catch was for: `--col <non-sqlite file>` → `"…: file is not a database"` envelope. Three new tests (forced, auto, CliRunner) plus the write-path twin. |
| 🟠 #2 stale title/body | ✅ Fixed | Title is `fix(direct):`; body has the three commit sections, an explicit **Behavior change** line, 1711/68 counts. Squash-merge will now carry the right record. |
| 🟡 `col?100%.db` Windows-illegal | ✅ | Parametrized `%` / `#` / space, `?` under `skipif(win32)` |
| 🟡 4 stripper vectors | ✅ | `<video>`, `<style type=…>\n…`, multi-line comment, `` all present |
| 🟡 dup test pins "still inserted" | ✅ (superseded) | Adapted to #61's raise semantics — `test_add_note_rejects_csum_duplicate_across_markup` pins refusal + nothing persisted + `allow_duplicate=True` still lands. The stronger pin. |
| 🟡 rebase | ✅ | Merged with `main`; base is `f3a0bfa`. The #59 fixture interaction (`update_note` → `_generate_missing_cards` needs `templates`/`cards`) and the `generated_cards` key are handled. The #60 argv test the conflict resolution dropped is back — net diff vs `main` on `test_app_bootstrap.py` is only the new CliRunner test. |

Head: 1711 passed, ruff clean, ty clean. ReDoS still fixed (2000 src-less quoted attrs → 0.6 ms); rslib vectors pass; strict entity decode unchanged. The regex/stripper block is byte-identical to round 2.

---

### 🟡 Nits

- **`anki_cli/db/anki_direct.py` `_ensure_write_safe`** (`f8c0be9`) — the rewrite lost the short-circuit. Round 2 had `_anki_process_running() or _sqlite_write_locked(...)`; now it's `running = …` then an unconditional `try: locked = _sqlite_write_locked(...)`. Verified: with `_anki_process_running → True`, the probe is still consulted (`['running', 'locked']`; round 2 stopped at `['running']`). So when Anki Desktop is open we now take a `BEGIN IMMEDIATE` on the live collection Anki has open — exactly what the spy test's docstring says the order is meant to avoid ("cheap process check before probing the live collection"). Outcome is a refusal either way, so no user-visible bug, but if the probe then errors the message becomes "Cannot verify…" instead of "Anki Desktop appears to be running", which is the less useful of the two. No test catches this; `test_write_refused_while_anki_process_running` leaves the probe real but doesn't assert it wasn't called.
  ```python
  if _anki_process_running():
      raise DirectWriteBlockedError(refusal)
  try:
      locked = _sqlite_write_locked(self.db_path)
  except sqlite3.Error as exc:
      raise DirectWriteBlockedError(f"Cannot verify … {exc}") from exc
  if locked:
      raise DirectWriteBlockedError(refusal)
  ```
  and in `test_write_refused_while_anki_process_running`, spy the probe and `assert consulted == ["running"]`.
- **`tests/integration/test_anki_direct_note_mutators.py:46`** — `_checksum()` is now dead. `main` (via #61) still used it in `_insert_note`; this PR switches `_insert_note` to `csum: int = 0` but the merge kept the helper and gave it a docstring. It's the line-for-line sha1 copy this PR exists to remove — delete it (no callers remain).

### 💡 Suggestion

- **`_probe_write_lock` message** (`detect.py`) — `f"Cannot probe the collection lock state at {db_path}: {exc}"` is fine; consider appending the remedy the other `DetectionError`s carry ("Check that --col points at a readable Anki collection file"), since this is the branch a user with a typo'd path will hit.

---

### ✅ Praise

- **Filing #62 and pointing `f8c0be9` at it** rather than leaving the fail-open bug under the #38 test-coverage issue — a bisect now lands on the right context. That was a 🟡 remark in round 2 and it got the proper fix, not a footnote.
- **The `DatabaseError` case is covered by design, not by accident.** The wrapper catches `sqlite3.Error` and the body explains why; the non-SQLite-file repro confirms it produces the envelope. Easy to have stopped at `OperationalError` and left "file is not a database" as the next traceback.
- **Merge-fallout audit.** Noticing that a conflict resolution silently dropped an unrelated test (#60's argv-sugar regression) and restoring it verbatim in its own commit is exactly the discipline a busy `main` needs; most merges of this size lose something and nobody checks.

---

*Unchanged from prior rounds: rslib/htmlescape parity claims are from reading upstream source, not executing rslib. Security/performance/architecture specialists were not re-run — the delta is the two boundary wrappers plus tests, and the ReDoS numbers were re-measured directly.*

umangk-2410:
Heads-up: `main` moved again 6 minutes after my approval — #52 (`6c9f3ed`) landed and the PR is now `CONFLICTING`. I did the merge locally to check what it takes; two conflicting files, plus two tests that pass the textual merge but fail because #52 changed the semantics underneath them. Full suite is green after the steps below (1772 passed, ruff clean).

### 1. `anki_cli/backends/detect.py` — two hunks in `detect_backend`

#52 replaced `_resolve_direct_collection(col_override)` + the `path is None` check with `_require_direct_collection(col_override, profile, empty_message=…)` on both the forced-direct and auto paths. Take **main's side** for both hunks, then re-apply this PR's one-token change on top:

```python
# forced-direct branch
        path = _require_direct_collection(
            col_override,
            profile,
            empty_message="Direct backend forced, but no Anki collection DB was found.",
        )
        if _anki_process_running() or _probe_write_lock(path):      # <- main has _sqlite_write_locked here

# auto branch
    direct_path = _require_direct_collection(
        col_override,
        profile,
        empty_message="No AnkiConnect and no collection found.",
    )
    if _anki_process_running() or _probe_write_lock(direct_path):   # <- same
```

`_probe_write_lock` itself merges cleanly (it sits below the conflict region). Sanity check after resolving: `grep -c "_probe_write_lock(" anki_cli/backends/detect.py` → `3` (the def + two call sites).

### 2. `tests/cli/test_app_bootstrap.py` — keep both sides

This PR's `test_cli_unopenable_collection_emits_backend_unavailable` and #52's `_raise_exit3` / `test_backend_free_commands_skip_detection` / `test_status_reports_*` were added at the same spot. Keep both blocks, two blank lines between.

### 3. Two semantic fallouts (merge cleanly, then fail)

- **`test_cli_unopenable_collection_emits_backend_unavailable`** invokes `version`. #52 added `_BACKENDLESS = {"version", "status", "config", "config:path", "config:set"}` — those commands now skip detection entirely, so the probe never runs and the test gets exit 0. Switch to a command that does detect, e.g. `"decks"`:
  ```python
  ["--backend", "direct", "--format", "json", "--col", str(bad), "decks"],
  ```
- **`test_auto_unopenable_collection_raises_detection_error`** stubs `_ankiconnect_reachable` as `lambda url: False`; #52 gave it a second parameter (`allow_non_localhost`), so the stub raises `TypeError` before the probe. Use `lambda *args, **kwargs: False`.

The existing forced-direct test and the write-guard tests are unaffected.

### Two nits from the round-3 review, if you're in there anyway

- `_ensure_write_safe` lost the `running or locked` short-circuit — it now takes `BEGIN IMMEDIATE` on the live collection even when Anki is running. Three-line reorder + `assert consulted == ["running"]` in `test_write_refused_while_anki_process_running`.
- `_checksum()` at the top of `test_anki_direct_note_mutators.py` is dead after `_insert_note` moved to `csum=0` — delete it.

Approval stands; none of this changes the verdict.

LouisDeconinck:
Rebased onto current main (conflict resolved — #52's `_require_direct_collection` refactor now drives collection resolution while the fail-closed `_probe_write_lock` wrapper is kept; the unopenable-collection e2e test moved from `version` to `decks` since `version` is now backendless). All 1772 tests pass, ruff clean.

PR #51, opened 2026-09-13
style: enable ruff W and apply whitespace fixes

LouisDeconinck:
Split out of the reworked #41/#46 cleanup per the review on #41: the whitespace/lint churn is unrelated to #33's scope and doubled the reviewable file count. This PR contains only the lint enablement and its mechanical fixes.

- Enable the `W` ruleset in ruff (W291/W292/W293) and run `ruff check . --fix`
- Lint the whole repo in CI (`ruff check .` instead of `ruff check anki_cli`, so `tests/` is covered)
- SIM118: replace `{k: row[k] for k in row.keys()}` with `dict(row)` — note `for k in row` would iterate *values* on `sqlite3.Row`, so `dict(row)` is the correct fix (not ruff's autofix shape)
- Split one overlong test signature (E501, now linted under `ruff check .`)
- Add `.DS_Store` to `.gitignore`

Gates: `ruff check .` clean, `pytest` 621 + 56 tui green.

Rebased onto current `main` (includes #40/#47), so there is no conflict with the anki_direct.py rewrites.


ubermenchh:
## Review

**Verdict: approve on content, but needs a rebase before merge** (see "Merge conflicts" below).

The PR does exactly what the description says. Whitespace-insensitive diff is 12 files / +25 −20; every `anki_cli/**` hunk is trailing-space or final-newline. The only semantic edit is the 17× `dict(row)` in integration-test helpers. I verified the stated gates on the PR head independently: `ruff check .` clean, `ty check anki_cli` clean, 621 + 56 tui tests green.

### Merge conflicts

GitHub reports `CONFLICTING` against `main`. #48 (`ced49d9`) landed after this branch was cut and touches 11 of the same files. A test merge shows two content conflicts:

- `tests/integration/test_anki_direct_answer_card.py`
- `tests/tui/test_browse_app.py`

Everything else auto-merges. Two things to watch on the rebase:

1. #48 introduced new `W` violations that weren't on `main` when this branch was cut (with this PR's `pyproject.toml` applied to current `main`: 41× W293, 35× W291). After rebasing, re-run `uv run ruff check . --fix` and re-commit, or CI's new `ruff check .` step will fail on the very rules this PR enables.
2. The W293 hits are `blank-line-with-whitespace`, a rule this PR hasn't had to fix yet, so expect a slightly larger whitespace diff than the current one.

### What I checked and found nothing wrong with

- **`dict(row)` vs `{k: row[k] for k in row.keys()}`**: identical for `sqlite3.Row` on 3.12/3.13. `dict()` uses the `keys()`/`__getitem__` mapping protocol, same key order, same duplicate-column handling. All 17 SELECTs are single-table with explicit non-repeating column lists, and every `fetchone` site has `assert row is not None` before the conversion. The 17 edits map 1:1 to the 17 SIM118 hits on `main`.
- **Rejecting ruff's autofix was correct**: `{k: row[k] for k in row}` iterates *values* on `sqlite3.Row`. With `ord == 0` it would silently return `{0: <id>, ...}` rather than raise. Good catch in the description.
- **Coverage**: consumers index the helper dicts by column name against literal values (`_card_row(...)["queue"] == -1` in `test_anki_direct_suspend_and_tags.py:125-134`, `[(int(row["ord"]), str(row["name"])) ...]` in `test_anki_direct_notetype_mutators.py:227,369`), so a wrong `dict(row)` would `KeyError` immediately. 621-green is real evidence, not incidental.
- **`ruff check .` scope**: repo root has only `anki_cli/` and `tests/` as Python trees; `.venv`/caches are default-excluded, `anki_cli/proto` stays in `extend-exclude`. Net new scope is `tests/`, as intended.
- **`W` ruleset**: no conflicts with the existing `E`/`RUF` selection. W191 is the only formatter-sensitive rule and only bites with `indent-style = "tab"`, which isn't set.

### Nit

- Single commit bundles reproducible tool output (75 W291/W292 files) with hand edits (pyproject, CI, `.gitignore`, `dict(row)`, E501 split). Since you're rebasing anyway, consider splitting into "config + hand fixes" and "mechanical `ruff --fix` output" so the second can go in a blame-ignore list without hiding the real changes. Preference, not a blocker.

### Follow-ups (not for this PR)

- **`.git-blame-ignore-revs`**: add the mechanical commit's sha post-merge so ~120 lines in `anki_direct.py`/`repl.py`/74 final-newline files don't blame to a style PR.
- **`.editorconfig`** with `trim_trailing_whitespace = true` / `insert_final_newline = true`: the 39 trailing-space lines in `anki_direct.py` are all on continuation-argument lines, and #40/#47/#48 all landed with them, so this is an editor setting, not one-off slips. Catching in CI is the expensive place.
- **Formatting is still unenforced**: `black>=26.1.0` is in `dev` deps with no `[tool.black]`, `[tool.ruff] line-length = 100`, and *neither* tool agrees with the tree (`ruff format --check`: 35 files on this branch, 85 on `main`; `black --check`: 71 files). Worth a separate mechanical PR in this same style: pick one (ruff format), drop black, run once, add `ruff format --check .` to CI. Might be worth noting in this PR's description that `W` does not mean "formatting is now enforced."
- **`ty check anki_cli`** still doesn't cover `tests/` while `ruff` now does. Probably deliberate; a one-line comment in the workflow would stop someone "fixing" it.
- **Row-helper duplication** (14 of the 17 rewritten helpers share the exact shape of `conftest.col_row()`): correctly left alone here. For #37, the thing to share is the connect/`row_factory`/close/`assert`/`dict(row)` scaffold as `fetch_one(db_path, sql, params)` / `fetch_all(...)`; keep each helper's explicit SELECT column list, since it doubles as a schema assertion.

LouisDeconinck:

PR #46, opened 2026-09-12
refactor: sweep dead code, duplicated defaults, and unused config fields (reopened #41)

ubermenchh:
Continuation of #41 by @LouisDeconinck, which was merged by mistake before its review was resolved and then reverted in #45. This PR is the original commit 9ca5cc7 (authorship intact), rebased onto nothing — `main` is back at the pre-#41 state, so the diff is identical to what #41 showed.

The review on #41 (changes requested) still applies in full; see https://github.com/ubermenchh/anki-cli/pull/41#pullrequestreview-5187733520. Short version of what needs to land before this merges:

**Blocking**
1. `CollectionConfig.path` still defaults to the deleted standalone path and `_serialize_config_toml` writes it on every `config:set`, which then acts as a missing `--col` override → every command exits 3 and the new `anki_profile` setting can't be enabled through the CLI. Fix: default `path` (and `anki_profile`) to `None`, skip `None` leaves when serializing.
2. Eager `detect_backend` in `app.py` now raises for `version` / `status` / `config:*` / bare `anki` when no Anki is present. Fix: skip detection for backend-free commands (or land #32 first).

**Important**
3. Unmatched `anki_profile` silently falls back to a different profile; default `"User 1"` means it's always set.
4. Split the `ruff W` whitespace autofix (~51 files) into its own PR; it conflicts with #40.
5. Disclose the undisclosed breaking changes (`config:path` keys, dropped TOML sections, `prefer = "standalone"` lockout, profile selection change).
6. Tests: capture the `anki_profile` kwarg in the detect stubs; add a backend-free-commands-survive-detection-failure test; add profile-not-found and `--col`-wins cases.
7. README/SKILL.md: document exit code 3 and `[collection] anki_profile`.

@LouisDeconinck — since this branch lives in the upstream repo you can't push to it directly. Either (a) open a fresh PR from your fork's `chore/issue-33-cleanup` once you've addressed the above and I'll close this one in favour of it, or (b) tell me and I'll add you as a collaborator on this branch. Apologies for the churn.

Closes #33 (when merged).

PR #52, opened 2026-09-13
feat!: rework dead-code & config sweep; fix config lockout and profile selection

### LINKED ISSUE #33: Dead code, duplicated defaults, unused config fields, repo hygiene

LouisDeconinck:
## What

Rework of the reverted #41 for the dead-code & config sweep in #33, rebuilt on current `main` and addressing the changes-requested reviews on #46 and this PR. **Supersedes #46.**

Closes #33.

The style-only work (ruff `W` enablement + whitespace/SIM118 fixes) was split out into #51 so this diff is functional only.

### Commits

- `refactor:` single-source shared defaults (`JSONValue` in `models/output.py`, `DEFAULT_ANKICONNECT_URL` in `models/config.py`), drop unused TUI symbols, consolidate shared helpers in `tui/_utils.py`
- `feat(config)!:` wire `collection.anki_profile` into detection; drop dead config fields
- `refactor!:` drop dead CLI surface (`init`, `--quiet`/`--verbose`, `status.message`) and the never-constructible `standalone` backend
- `docs:` exit code 3 + `[collection] anki_profile` in README/SKILL
- `fix!:` bare `anki` REPL detects its backend again; stale `main`-written config values are migrated at load; `status` reports real detection state

### Fixes from the reviews

- **Bare REPL was dead (blocking):** `None` was in `_BACKENDLESS`, so `run_repl` got `backend="none"` — a name the factory can't construct — and every REPL command failed. Now bare `anki` runs `detect_backend`: on success the REPL gets the real backend; on `DetectionError` it opens anyway with a stderr warning and `backend="none"`, so commands surface `BACKEND_UNAVAILABLE` per invocation instead of the shell refusing to start. Tests pin both paths plus factory-constructibility of the handed-over context.
- **Config lockout not cured for existing users (blocking):** the serializer fix only helped files written *from now on*. `load_app_config` now strips the stale marker `collection.path = "~/.local/share/anki-cli/collection.db"` — and, only when that marker identifies a `main`-written file, the never-read `anki_profile = "User 1"` default — warning via `meta.warnings`. The next `config:set` rewrites the file clean. A genuine `anki_profile = "User 1"` (no stale path marker) is preserved.
- **Unmatched profile killed a reachable AnkiConnect:** collection/profile lookup on the AnkiConnect branches is now informational — discovery (`_discover_collections` + `_pick_collection`) never raises; only the direct branches go through `_require_direct_collection`, which exits 3 naming the profile and listing available ones. An explicit missing `--col`/`collection.path` fails loudly naming the resolved path on direct paths, and is ignored on AnkiConnect paths.
- **`status` reported nothing:** it no longer reads the bootstrap context. It runs `detect_backend` itself and returns `{"ok", "backend", "collection", "reason", "profile"}` on success or `{"ok": false, "error": ...}` on failure — exit 0 either way, so it works as a health check. `--backend`/`--col` affect it again.
- **Warnings are structured:** stale config/env notices are collected in `LoadedConfig`/`RuntimeConfig.warnings` and emitted through `meta.warnings` instead of raw stderr lines, so `--format json` consumers parsing the stderr error envelope never hit a bare text line. Table output still echoes them as `warning:` lines.
- **Lockout-proof `config:set`:** `backend.prefer`/`display.default_output` are `Literal`-typed (all four allowed-value sets — CLI `Choice`, `_ALLOWED_*`, detection — now derive from `get_args`), so a bad value is rejected at write time instead of bricking the next run. `config:set <optional> ""` clears the field; `collection.path = ""` can no longer resolve to cwd. Values are serialized through a full TOML basic-string escaper (C0 → `\uXXXX`) and the output is re-parsed with `tomllib` before `os.replace`, so `config:set` can't persist an unparseable file.
- **Plumbing pinned:** `anki_profile=` is captured and asserted at the bootstrap and resolver call sites; near-miss names (`"User"`, `"ser 1"`, `"User 1 (copy)"`, `"user 1"`) pin the exact-match comparison. `_ankiconnect_reachable` honours `allow_non_localhost` — no probe leaks to a remote host that data ops would refuse anyway.
- **Smaller items:** `_coerce_raw_value` deleted (pydantic lax mode handles coercion); `_reason()` deleted in favour of a structured `DetectionResult.profile`; `anki_direct.py` imports `JSONValue` directly (no `TYPE_CHECKING` hack); utility tests moved to `tests/unit/test_tui_utils.py` so they run on every push; bootstrap tests no longer touch the real `$HOME` config.

## Breaking changes

- `config:path` JSON output drops `backups` and `standalone_default`.
- `backup.*`, `review.*`, and `display.day_boundary_hour` config keys no longer exist; the next `config:set` rewrites `config.toml` without those sections.
- `prefer = "standalone"` / `ANKI_CLI_BACKEND=standalone` warn (via `meta.warnings`) and run as `auto`; `--backend standalone` is rejected by the CLI with a usage error (exit 2).
- `init` command and `--quiet`/`--verbose` flags removed; the `status` payload drops `message` and now carries `ok`/`reason`/`profile`/`error` — it exits 0 even when no backend is found.
- Auto mode exits 3 when no backend is found. `collection.anki_profile` is now live: multi-profile installs may resolve a different collection than before — set `--col`/`collection.path` to pin a collection. Stale `main`-written `collection.path`/`anki_profile = "User 1"` defaults are ignored at load (see above).

## Validation

- `pytest` — 701 passed; `pytest -m tui` — 52 passed
- `ruff check anki_cli` and `ty check anki_cli` — clean
- Manual: bare `anki` against a stub AnkiConnect opens the REPL on `ankiconnect` and `decks` lists decks; bare `anki` on a host with no Anki opens the REPL with a stderr warning and `decks` reports `BACKEND_UNAVAILABLE`; `anki --format json status` exits 0 with `data.ok=false` and clean-JSON stderr; a `main`-shaped config file loads with stale values ignored and `config:set` rewrites it clean.

Note: pointing the direct backend at a 0-byte `collection.anki2` still surfaces a raw `sqlite3.OperationalError` traceback — identical behaviour on `main`, unchanged here.


ubermenchh (changes requested):
# Code Review — PR #52: dead-code & config sweep; config lockout + profile selection

**Verdict:** BLOCK (request changes)
**Scope:** `chore/issue-33-cleanup` → `main`, 4 commits, 34 files, +497/−512. Reviewed at HEAD `e58c181` on base `12934aa`.
**Counts:** 🔴 1 blocking · 🟠 5 important · 🟡 6 nit · 💡 7 suggestion · ✅ 4 praise

Verified locally in a throwaway worktree: `pytest` 628 passed / 5 skipped, `ruff check` clean. Every 🔴/🟠 below was reproduced by executing code, not just by reading.

## Summary

The cleanup half of this PR is well done: the `standalone` backend, `init`, `--quiet/--verbose`, dead config sections, and duplicated TUI helpers are gone with no collateral, and the serializer fix (derive sections from the model, skip `None`) removes the root cause of the lockout for files written *from now on*. Two things stop it from shipping.

1. Adding `None` to `_BACKENDLESS` hands the REPL `backend="none"`, which the factory cannot construct, so bare `anki` can no longer run a single backend command; the test was edited to assert the broken state.
2. The PR claims to fix the config lockout but does not migrate the files `main` already wrote: every existing user who ever ran `config:set` has `path = "~/.local/share/anki-cli/collection.db"` **and** `anki_profile = "User 1"` persisted, and this branch honours both, so those users still exit 3 and `config:set` re-writes the stale path.

Recommended next step: drop `None` from `_BACKENDLESS` (or make detection lazy in the factory), add a load-time legacy-value shim, make the informational AnkiConnect-branch profile lookup non-fatal, then re-request review.

---

## 🔴 Blocking

### 🔴 Bare `anki` REPL cannot execute any backend command
**`anki_cli/cli/app.py:29,131-138,165-168`** → **`anki_cli/tui/repl.py:196,287,323,806`** → **`anki_cli/backends/factory.py:48`** — `None ∈ _BACKENDLESS` skips detection and hands `run_repl` a context with `backend="none"`; there is no lazy resolution anywhere.

**Trace:** `app.py:131` `invoked_subcommand is None` → `:135` `backend="none"` → `:168` `run_repl(ctx.obj)` → `repl.py:287` `_fetch_due_counts` → `factory.py:19` `backend_name="none"` → `:48` `BackendFactoryError("Unknown backend 'none'.")` (swallowed; header shows no counts). User types `decks` → `repl.py:196` copies the same dict into the subcommand context → same raise → `BACKEND_UNAVAILABLE`, exit 7. `review`/`cards` hit the same path at `:323`/`:806`. `grep detect_backend anki_cli/tui/` returns nothing.

**Counterexample:** `anki` with a working AnkiConnect, then `decks`. Actual: `Unknown backend 'none'.` Expected (and `main`'s behaviour): deck table.

Reproduced: `backend_session_from_context({"backend": "none", "collection_path": None})` → `Unknown backend 'none'.`

**Tests:** `tests/cli/test_app_bootstrap.py:302-315` was rewritten to assert `backend == "none"` with the comment "per-command ops resolve lazily", and a `fail_detect` stub (`:302-305`) now *forbids* the correct behaviour. `run_repl` is monkeypatched to a recorder, so nothing checks that the handed-over dict is constructible.

**Structure:** a hardcoded name-set in `app.py` duplicates the knowledge "does this command call `backend_session_from_context`?", and the REPL is one subcommand (`None`) that runs many commands, so a per-command flag would not fix it either. The only seam where the answer is always right is the factory.

**Disclosure:** the description says "bare `anki` enters the REPL with backend 'none'" but not that this makes the REPL non-functional.

**Suggested fix** (smallest correct diff):
```python
# anki_cli/cli/app.py
_BACKENDLESS = {"version", "status", "config", "config:path", "config:set"}  # drop None

    if ctx.invoked_subcommand in _BACKENDLESS:
        ctx.obj.update({"collection_path": runtime.collection_override,
                        "backend": "none", "backend_reason": "not required"})
    else:
        try:
            detection = detect_backend(...)
        except DetectionError as exc:
            if ctx.invoked_subcommand is None:
                # REPL still opens on a host with no Anki; commands surface the error.
                click.echo(f"warning: {exc} (backend commands unavailable)", err=True)
                ctx.obj.update({"collection_path": runtime.collection_override,
                                "backend": "none", "backend_reason": str(exc)})
            else:
                formatter_from_ctx(ctx).emit_error(...)
                raise click.exceptions.Exit(exc.exit_code) from exc
        else:
            ctx.obj.update({...detection...})
```
Then revert `test_no_subcommand_runs_repl_when_available` / `test_no_subcommand_import_error_falls_back_to_help` to a stubbed *successful* detection and add a constructibility assertion:
```python
backend = factory_mod.create_backend_from_context(calls["obj"])  # raises on PR head
assert backend.collection_path == db
```
Structural alternative (~−15 net lines): move `detect_backend` into `factory.create_backend_from_context` behind `if obj.get("backend") in (None, "none")`, carry `exit_code` on `BackendFactoryError`, delete `_BACKENDLESS` and `app.py:131-163` entirely. That makes the REPL correct by construction and is what #32 wants anyway.

---

## 🟠 Important

### 🟠 Migrate the stale `collection.path` that `main` persisted; the lockout is not cured for existing users
**`anki_cli/config_runtime.py:236-237` (honours it), `:129-143` (re-writes it)** — `main`'s `_serialize_config_toml` wrote every field, so any user who ever ran `config:set` has `path = "~/.local/share/anki-cli/collection.db"` on disk. This branch still treats it as an explicit `--col` to a nonexistent file.

**Trace:** `load_app_config` → `file_data["collection"]["path"]` present and non-`None` → `:236-237` returns it as `collection_override` → `detect.py:78` → `:123-125` `exists()` False → `None` → `:96-99` exit 3 `"No AnkiConnect and no collection found."` (message never names the path). User runs `config:set collection.anki_profile Work` to recover: `:130` `model_dump` includes the stale path → `:341` not `None` → **re-written**.

Reproduced with a `main`-shaped file: after `config:set`, the file still contains `path = "~/.local/share/anki-cli/collection.db"` and `_resolve_collection_override` returns it.

The PR maps `prefer = "standalone"` for exactly this reason (`:168-174`) but not the far more common stale path. `tests/unit/test_config_runtime.py:251-301` only covers a fresh file.

**Suggested fix:**
```python
# config_runtime.py
_LEGACY_STANDALONE_PATH = "~/.local/share/anki-cli/collection.db"

def load_app_config(config_path=None) -> LoadedConfig:
    ...
    col = parsed.get("collection")
    if isinstance(col, dict) and col.get("path") == _LEGACY_STANDALONE_PATH:
        sys.stderr.write("warning: dropping legacy collection.path left by the removed standalone backend\n")
        del col["path"]
    ...
# detect.py: when col_override is set but missing, fail loudly instead of falling through to "no collection found":
#   raise DetectionError(f"Collection override {col_override} does not exist.", exit_code=3)
```
Red-on-HEAD test:
```python
def test_set_config_value_drops_legacy_dead_default_path(tmp_path):
    cfg = tmp_path / "config.toml"
    cfg.write_text('[collection]\npath = "~/.local/share/anki-cli/collection.db"\n')
    set_config_value(key="collection.anki_profile", raw_value="Work", config_path=cfg)
    assert "collection.db" not in cfg.read_text()
```

### 🟠 Persisted `anki_profile = "User 1"` default from `main` is now a hard filter
**`anki_cli/backends/detect.py:145-153`** — `main`'s `CollectionConfig.anki_profile` defaulted to `"User 1"` and was persisted by every `config:set`, but never read. This PR makes it live, so any existing user whose profile directory is not literally `User 1` exits 3 on every backend command, in all three detection branches.

Reproduced: file with `anki_profile = "User 1"`, real profile `Umang` → `Anki profile 'User 1' not found. Available: Umang.` exit 3. The breaking-change note ("multi-profile installs may resolve a different collection") understates this: single-profile installs with a non-default name break outright.

**Suggested fix:** fold into the legacy shim above; a `main`-written file is identifiable by the stale path marker:
```python
if isinstance(col, dict) and col.get("path") == _LEGACY_STANDALONE_PATH:
    col.pop("path", None)
    if col.get("anki_profile") == "User 1":
        col.pop("anki_profile", None)   # never user-chosen on main
```
Add this to the breaking-change list explicitly.

### 🟠 Unmatched profile aborts a reachable AnkiConnect session
**`anki_cli/backends/detect.py:48-52,71-76`** — in both AnkiConnect branches `_resolve_direct_collection(..., anki_profile=)` only populates a display-only `collection_path`, yet its new raise at `:150-153` propagates and exits 3 even though the backend just answered `version`.

**Counterexample:** profiles `["User 1"]`, `anki_profile="Work"`, AnkiConnect reachable, `anki --backend ankiconnect decks` → exit 3 `Anki profile 'Work' not found`. On `main`: `DetectionResult("ankiconnect", None, "forced")` and a deck listing. Behaviour is also inconsistent: with no local `Anki2` directory at all (remote AnkiConnect, README-supported) `:142-143` returns `None` and nothing raises, so success depends on unrelated local state.

Reproduced: `detect_backend(forced_backend="ankiconnect", anki_profile="Work")` with reachable stub → exit 3.

A `-> Path | None` helper that now raises is a leaky contract; the decision belongs in `detect_backend`'s direct branches only. Untested either way: every `detect_backend` test replaces `_resolve_direct_collection` with a non-raising spy (`tests/backends/test_detect.py:215-241`).

**Suggested fix:** split discovery from the decision:
```python
def _discover_collections(col_override) -> list[Path]: ...   # lines 123-140, pure
def _pick_collection(candidates, anki_profile) -> Path | None:
    if not anki_profile:
        return candidates[0] if candidates else None
    return next((p for p in candidates if p.parent.name == anki_profile), None)
def _require_direct_collection(col_override, anki_profile) -> Path:  # raises exit 3 with "Available: ..."
# ankiconnect branches: _pick_collection(_discover_collections(...), profile)
# direct branches:      _require_direct_collection(...)
```
And pin it:
```python
def test_forced_ankiconnect_with_unmatched_profile_still_succeeds(monkeypatch, tmp_path):
    # real _anki_data_roots with ["Personal"], _ankiconnect_reachable=lambda url: True
    result = detect_backend(forced_backend="ankiconnect", anki_profile="Work")
    assert result.backend == "ankiconnect" and result.collection_path is None
```

### 🟠 `status` can no longer report status
**`anki_cli/cli/commands/general.py:33-47`**, **`README.md:33-38,78`**, **`SKILL.md:13`** — `status` is in `_BACKENDLESS`, so `data.backend` is the constant `"none"` and `data.collection` is only the explicit `--col`. README line 78 shows `anki --backend direct --col … status` to verify a collection; `--backend` is now ignored for that command. Docs, docstring ("Show current backend and collection status"), and the SKILL.md agent contract are all wrong, and this is not in the breaking-change list. `tests/cli/test_general_commands.py` misses it because it injects a fake `backend="ankiconnect"` obj.

**Suggested fix:** `status` is the one command that should run detection and turn a `DetectionError` into *data*, not an exit code:
```python
try:
    detection = detect_backend(forced_backend=obj["requested_backend"],
                               col_override=obj.get("collection_override"),
                               ankiconnect_url=cfg.backend.ankiconnect_url,
                               anki_profile=cfg.collection.anki_profile)
    data = {"ok": True, "backend": detection.backend,
            "collection": str(detection.collection_path) if detection.collection_path else None,
            "reason": detection.reason}
except DetectionError as exc:
    data = {"ok": False, "backend": None, "collection": None, "error": str(exc)}
formatter.emit_success(command="status", data=data)   # exit 0 either way
```
Remove `"status"` from `_BACKENDLESS` and update README/SKILL.md accordingly.

### 🟠 The headline "config lockout" fix has no end-to-end test, and the one test that exists reads the developer's real HOME
**`tests/cli/test_app_bootstrap.py:118-126`** — `test_backend_free_commands_survive_detection_failure` parametrizes `version`/`status`/`config:path` only. `config` and `config:set` (the command a locked-out user actually needs) are unpinned; a typo in the `"config:set"` literal passes CI. It is also the only bootstrap test that leaves `resolve_runtime_config` live, so it reads `~/.config/anki-cli/config.toml`; adding `config:set` to the list naively would *write* there. Surviving mutations: drop `"config"`/`"config:set"` from the set; change `:134`/`:135` values.

**Suggested fix:**
```python
@pytest.mark.parametrize("argv", [["version"], ["status"], ["config"], ["config:path"],
                                  ["config:set", "--key", "display.color", "--value", "false"]])
def test_backend_free_commands_skip_detection(monkeypatch, tmp_path, argv):
    runtime = _runtime(backend="auto", output_format="json",
                       collection_override=tmp_path / "override.anki2")
    runtime.config_path = tmp_path / "config.toml"        # keep config:set off the real disk
    monkeypatch.setattr(app_mod, "resolve_runtime_config", lambda **kw: runtime)
    monkeypatch.setattr(app_mod, "detect_backend", _raise_exit3)
    result = CliRunner().invoke(app_mod.main, ["--format", "json", *argv])
    assert result.exit_code == 0, result.output
    payload = json.loads(result.output)
    assert payload["ok"] is True
```

---

## 🟡 Nits

- **`anki_cli/config_runtime.py:133-143,167-177`** — `config:set backend.prefer direkt` (or `standalone`, or `display.default_output xml`) succeeds because the pydantic fields are bare `str`; the next run exits 2 in `_resolve_backend` *before* the `_BACKENDLESS` gate, so even `config:set` is locked out (escape hatch `--backend auto` is undocumented). Mostly pre-existing on `main`; the PR adds the `standalone` warn-forever variant. Fix: make `prefer`/`default_output` `Literal[...]` on the model (`models/config.py:16,22`) and derive `_ALLOWED_BACKENDS`/`_ALLOWED_OUTPUTS`/`click.Choice`/`detect.py:36` from `get_args(...)`. That also collapses four hand-maintained copies of the same set.
- **`anki_cli/config_runtime.py:236-237,305-309`** — `config:set collection.path ""` persists `path = ""`; `Path("").resolve()` is `cwd`, which detection accepts and `DirectBackend` then fails on with a raw `sqlite3.OperationalError`. `ANKI_CLI_COLLECTION=""` is already rejected at `:232-233`; do the same, or treat `""` as unset for optional fields (`return stripped or None`).
- **`anki_cli/backends/detect.py:147`** — exact `==` against `path.parent.name` while `config:set` stores the raw string; `"Work "` never matches `Work`. Use `anki_profile.strip()`.
- **`tests/tui/test_utils.py:5-9`** — `importorskip("textual")` + `pytest.mark.tui` for a module that imports only `time`. `pyproject.toml` has `addopts = "-m 'not tui'"` and `tui.yml` only runs on path-filtered PRs, so this test never runs on push to `main`. Move to `tests/unit/` and drop both guards.
- **`anki_cli/config_runtime.py:168-174` / `tests/unit/test_config_runtime.py:164-187`** — only the file path is tested; `ANKI_CLI_BACKEND=standalone` is claimed in the description but unexercised (low risk, same `candidate` collapse). Also `--backend standalone` is rejected by `click.Choice` with a usage error, not the friendly warning; say so in the description/README migration note.
- **PR title** — the squash title is `refactor!:` but the PR adds user-visible behaviour (`anki_profile`, exit 3); `feat!:` is what changelog tooling expects.

## 💡 Suggestions

- **`anki_cli/config_runtime.py:350-359`** — `_toml_scalar` escapes only `\` and `"`. A value with `\n` cannot inject keys (tomllib rejects raw control chars in basic strings) but *does* produce an unparseable file → exit 2 on every command including `config:set`, the exact lockout the PR is fighting. Escape C0 chars as `\uXXXX` and `tomllib.loads(content)` before `os.replace` so you never persist something you can't read back.
- **`anki_cli/config_runtime.py:171-173`** — the warning fires on every invocation forever (`prefer` stays `"standalone"` in the model and is re-written by every `config:set`) and doesn't say how to clear it. Append: `Run: anki config:set --key backend.prefer --value auto`. Better: return warnings from `resolve_runtime_config` and route them through `Meta.warnings` (`models/output.py:19`) instead of a raw `sys.stderr.write` from a non-CLI module; `--format json` consumers reading stderr for the error envelope now get a non-JSON line first.
- **`anki_cli/backends/detect.py:112-116`** — `_reason()` formats a string nobody reads: `grep backend_reason anki_cli/` finds two writers and zero readers, and the profile that won is already `collection_path.parent.name`. Delete the helper and its four wraps; change the test to assert `result.collection_path.parent.name == "Work"`. If `status` starts displaying `reason` (per the 🟠 above), keep it but add a `profile: str | None` field rather than parsing prose. Either way the `col_override is not None` branch is currently unpinned.
- **`anki_cli/config_runtime.py:305-319`** — `_coerce_raw_value` encodes model knowledge in comments that will rot. Less code: delete it and pass `raw_value` straight to `AppConfig.model_validate` at `:139`; pydantic lax mode already coerces `"true"/"false"/"yes"/"no"/"on"/"off"/"1"/"0"` and rejects strings assigned to a nested model (the section-key case). Only the error text changes.
- **`anki_cli/backends/detect.py:101-110`** — `_ankiconnect_reachable` POSTs to whatever URL is configured without honouring `allow_non_localhost`; `AnkiConnectBackend._validate_url` gates data ops, so nothing unsafe is reachable, but one probe per invocation leaks to a remote host. Pre-existing; this PR narrows it by not probing for `config*`/`version`. Thread the flag through when convenient.
- **`anki_cli/tui/_utils.py`** — right home, but finish the job: `_to_int`/`_safe_int` (`browse_app.py:63-67` / `review_app.py:81-85`) are byte-identical, and the `review_day_index` "today/tomorrow/Nd" arm is duplicated verbatim next to the `relative_eta` call you just extracted.
- **`tests/backends/test_detect_helpers.py:160-172`** — a `==` → `in`/`.startswith` mutation at `detect.py:147` survives; add near-misses `"User"`, `"ser 1"`, `"User 1 (copy)"` to the unmatched test.

## ✅ Praise

- **`tests/unit/test_config_runtime.py:251-301`** — `test_set_anki_profile_does_not_persist_collection_path` walks serialize → reload → `_resolve_collection_override` → `_resolve_direct_collection`. Because `_toml_scalar(None)` raises, the `None`-skip at `:341-342` is genuinely exercised rather than incidentally passing. Right shape for a data-flow regression test; it just needs the stale-file sibling.
- **`tests/backends/test_detect.py:215-243`** — one parametrized test pins all four `anki_profile=` call sites (`detect.py:50,55,74,78`); that is exactly the "plumbing pinned" guarantee the description promises. `test_detect_helpers.py:175-188` kills the `.lower()`-both-sides mutation with `"user 1"` vs `"User 1"`, and `test_utils.py:22` (`36h → "2d"`) is the single row that distinguishes ceiling from floor division.
- **`anki_cli/backends/protocol.py:6`, `anki_cli/models/output.py:8-9`** — `JSONValue` moved to the leaf layer in the right direction: `models/` imports nothing from `anki_cli`, so `backends → models` is acyclic and adds no startup imports. Free follow-up: `anki_cli/db/anki_direct.py:19-22`'s `TYPE_CHECKING` / `JSONValue = Any` hack existed only to dodge a `db → backends → db` cycle and can now be a plain import.
- **`anki_cli/cli/app.py:131-138`**, **`anki_cli/config_runtime.py:331-347`** — `anki version` / `config:set` no longer pay for an httpx POST (up to 0.35 s against a down remote), a `pgrep`/`tasklist` fork (2–3 s timeout), a `/proc` walk, and a `BEGIN IMMEDIATE`. Deriving TOML sections from `model_dump()` also removes a hardcoded list that was already stale. The `standalone`/`init`/`--quiet` removals are the good kind of diff: less surface, fewer lies in `--help`, no formatting churn across 34 files.

## Notes on method

- Two of the findings above are "test edited to match the code". A `mutmut` run scoped to `anki_cli/cli/app.py`, `anki_cli/backends/detect.py`, `anki_cli/config_runtime.py` (~640 lines, fast hermetic tests once the real-HOME dependency is removed) would give cheap real signal.
- Commit-by-commit buildability (`git rebase -x pytest`) was not run; TUI tests were skipped in the review env (textual not installed), so `pytest -m tui` was taken on trust.

LouisDeconinck:

LouisDeconinck:

LouisDeconinck:
All findings addressed in `2358be5` (pushed), plus PR title → `feat!:` and the description updated to match. Validation: `pytest` 701 passed, `pytest -m tui` 52 passed, `ruff check anki_cli` + `ty check anki_cli` clean.

**🔴 Bare `anki` REPL** — done per the suggested fix: `None` dropped from `_BACKENDLESS`, bootstrap wraps `detect_backend` and on `DetectionError` echoes `warning: ... (backend commands unavailable)` to stderr and degrades to `backend="none"` for the REPL only — `run_repl` opens, commands surface `BACKEND_UNAVAILABLE` via the factory (which now reports the recorded detection reason instead of `Unknown backend 'none'`). `test_no_subcommand_runs_repl_when_available` reverted to a stubbed successful detection + a `create_backend_from_context(calls["obj"])` constructibility pin; a new test covers the degraded-with-warning path.

**🟠 Stale `collection.path` / `anki_profile = "User 1"`** — `_normalize_legacy_values` runs inside `load_app_config`: the stale standalone-path marker is dropped (and the `"User 1"` default with it, only when the marker identifies a `main`-written file — a genuine `User 1` profile survives), with a warning naming the file. `config:set` rewrites the file clean. A missing explicit `--col`/`collection.path` now fails loudly: `Collection override <resolved path> does not exist.` (exit 3).

**🟠 Profile miss kills reachable AnkiConnect** — split exactly as suggested: `_discover_collections` (pure), `_pick_collection` (informational, never raises), `_require_direct_collection` (raises exit 3 listing `Available: ...`). Both AnkiConnect branches use the informational path — a profile miss or stale `--col` can't veto a backend that just answered `version`. Pinned by `test_forced_ankiconnect_with_unmatched_profile_still_succeeds` and the `auto` variant.

**🟠 `status`** — removed from the context-reading path entirely: it calls `detect_backend` itself and returns `{"ok", "backend", "collection", "reason", "profile"}` on success or `{"ok": false, "backend": null, "collection": null, "error"}` on failure — exit 0 either way, so `--backend`/`--col` work with it again. README + SKILL.md updated.

**🟠 Bootstrap test** — parametrizes `version`/`status`/`config`/`config:path`/`config:set`, pins `config_path` to `tmp_path`, and `status`'s own `detect_backend` reference is stubbed separately.

**🟡 nits** — `prefer`/`default_output` are `Literal`-typed; `_ALLOWED_BACKENDS`/`_ALLOWED_OUTPUTS`, both `click.Choice`s, and `detect.py`'s forced-backend check all derive from `get_args`. Empty-string optional values clear the field (`collection.path = ""` → unset, never cwd). `anki_profile.strip()` before matching (plus a dedicated whitespace test). Utility tests moved to `tests/unit/test_tui_utils.py` with the `tui` mark and `importorskip` dropped. `ANKI_CLI_BACKEND=standalone` is exercised. README migration note now says `--backend standalone` is rejected at the CLI.

**💡 suggestions** — `_toml_scalar` escapes C0/DEL as `\uXXXX` and `_write_config_file` re-parses with `tomllib` before `os.replace`. Warning text now says how to clear it, and — taking the "Better:" — warnings ride `LoadedConfig`/`RuntimeConfig.warnings` into `meta.warnings` (verified: `--format json` error envelope on stderr stays pure JSON). `_reason()` deleted; `DetectionResult.profile` added and surfaced in `status`. `_coerce_raw_value` deleted — pydantic lax mode coerces/validates. `allow_non_localhost` threaded into `_ankiconnect_reachable` — no probe leaks to a refused remote host. `_to_int`/`_safe_int` and the duplicated due-label arm consolidated as `to_int`/`due_day_label` in `tui/_utils.py`. Near-miss profile names (`"User"`, `"ser 1"`, `"User 1 (copy)"`) added to the unmatched-profile test. `anki_direct.py` imports `JSONValue` directly.

**Manual verification** — bare `anki` against a stub AnkiConnect opens on `ankiconnect` and `decks` lists decks; bare `anki` with no Anki opens the REPL with the warning and `decks` reports `BACKEND_UNAVAILABLE`; `anki --format json status` on a no-Anki host exits 0 with `data.ok=false`; a `main`-shaped `config.toml` (stale path + `User 1` + `prefer = "standalone"`) loads with both warnings in `meta.warnings` and `config:set` rewrites it clean.

The `git rebase -x pytest` / mutmut suggestions are noted but not run — out of scope for this pass.

ubermenchh:
**Merge conflicts with `main`** — the PR is currently `CONFLICTING` (`mergeStateStatus: DIRTY`) after #49, #53 and #54 landed. I did a trial merge of `origin/main` (`3c03d64`) into `2358be5`: only **two hunks** conflict, both mechanical. Everything else — including #49's changes to `browse_app.py` / `review_app.py` / `repl.py` that overlap the TUI consolidation here — auto-merges cleanly, and `main` introduced no new references to any symbol this PR removes (`BackendNotImplementedError`, `_relative_eta`, `_to_int`, `_safe_int`, `standalone`, …), so nothing should break after resolution.

### 1. `anki_cli/backends/factory.py` — import block

#54 added `UnsupportedCollectionError`; this PR added `DEFAULT_ANKICONNECT_URL`. Keep both:

```python
from anki_cli.db.anki_direct import UnsupportedCollectionError
from anki_cli.models.config import DEFAULT_ANKICONNECT_URL, AppConfig
```

#54's `except (FileNotFoundError, UnsupportedCollectionError) as exc:` on the `DirectBackend` construction auto-merges — leave it. The `if backend_name == "standalone": raise BackendNotImplementedError(...)` block that still exists on `main` is correctly deleted by this PR's side; the new `if backend_name in {"", "none"}:` branch should sit where it does now.

### 2. `anki_cli/cli/commands/review.py` — `review:answer` `except` tuple

#49 inserted the revlog-id undo block (`if snapshot is not None: ... UndoStore().push(undo_item)` with the `OSError` warning) immediately above an `except` clause that this PR rewrote. Keep **#49's undo block** and **this PR's `except` tuple**:

```python
            if snapshot is not None:
                # Carry the id of the revlog row this answer wrote so undo can
                # delete exactly that row instead of a time window.
                revlog_id = result.get("revlog_id") if isinstance(result, Mapping) else None
                undo_item = UndoItem(
                    collection=collection,
                    card_id=int(card_id),
                    snapshot={**snapshot, "revlog_id": revlog_id},
                    created_at_epoch_ms=now_epoch_ms(),
                )
                # Best-effort: the answer is already committed, so a failed
                # undo write must not fail the command (a retry would apply a
                # second review).
                try:
                    UndoStore().push(undo_item)
                except OSError as exc:
                    click.echo(
                        f"warning: answer saved but undo entry not written: {exc}",
                        err=True,
                    )
    except (BackendFactoryError, NotImplementedError) as exc:
```

i.e. drop `BackendNotImplementedError` from the tuple as everywhere else in this PR.

### After resolving

```sh
git rebase origin/main            # or merge, per repo convention
uv run ruff check anki_cli        # will catch a leftover BackendNotImplementedError name if the tuple is mis-resolved
uv run pytest -q
uv run pytest -q -m tui
```

A quick `rg -n 'BackendNotImplementedError|standalone' anki_cli tests` should return only the legacy-value shim in `config_runtime.py` and its tests.

LouisDeconinck:
Rebased onto `main` @ `f3a0bfa` (#49, #53–#56, #59–#61 all included) — `mergeStateStatus` is back to BLOCKED (clean merge), head is now `995540b`.

The two hunks you flagged resolved exactly as suggested:

- **`factory.py`** — kept both imports (`UnsupportedCollectionError` + `DEFAULT_ANKICONNECT_URL`); the `standalone`/`BackendNotImplementedError` block stays deleted, and the `{"", "none"}` degraded-backend branch sits where the standalone arm was.
- **`review.py`** — kept #49's revlog-id undo block verbatim and this PR's `except (BackendFactoryError, NotImplementedError)` tuple.

Rebasing (rather than merging) surfaced a few more overlaps the trial merge couldn't show, all resolved:

- **`note.py`** (`note:bulk`) — #61's `add_notes(..., allow_duplicate=allow_duplicate)` kept, with the slimmed `except` tuple.
- **`anki_direct.py`** — kept the new `core.template` (#59) and `db.timing` (#55) imports; the `TYPE_CHECKING` JSONValue shim is replaced by the direct `models.output.JSONValue` import (the `type`-alias single-sourcing makes the guard unnecessary).
- **`browse_app.py` / `review_app.py`** — one subtlety worth flagging: `_days_from_today` (#55) landed in the exact region where `_relative_eta`/the duplicated today/tomorrow arm were removed. The consolidated `due_day_label` in `tui/_utils.py` now **absorbs `_days_from_today`'s semantics** — it prefers the backend's `days_from_today` count and treats `epoch_secs` as the *start* of the due day (ceil, so a rollover 21 h away is still "tomorrow"). No regression of #55's fix; `tests/unit/test_tui_utils.py` pins both the ceiling rows and `days_from_today` preference.
- **`tests/tui/test_browse_app.py`** — both `browse_mod.time` monkeypatch targets repointed at `utils_mod.time` (browse_app no longer imports `time`).

Post-resolution sweep per your checklist: `rg -n 'BackendNotImplementedError|standalone' anki_cli tests` returns only the legacy-value shim in `config_runtime.py` and its tests. Validation: `pytest -q` → **1728 passed / 63 deselected**, `pytest -q -m tui` → **63 passed**, `ruff check anki_cli tests` clean.

Ready for re-review whenever you are.

PR #54, opened 2026-09-13
fix(direct): refuse collections older than schema 18 with an upgrade hint

### LINKED ISSUE #24: No schema-version guard: legacy schema-11 collections crash with opaque errors
## Problem

`AnkiDirectReadStore.__init__` (`anki_cli/db/anki_direct.py:35-42`) only checks that the file exists. The whole module assumes schema ≥ 18 (separate `decks`, `notetypes`, `fields`, `templates`, `deck_config` tables with protobuf blobs).

Legacy schema-11 collections (Anki ≤ 2.1.27, or any profile never opened in a modern Anki) store decks/models/dconf as JSON in the `col` table and have none of those tables. `backends/detect.py:118` happily picks up such a `collection.anki2`.

## Reproduce

Profile from Anki 2.1.26:

```sh
anki decks
# sqlite3.OperationalError: no such table: decks
```

Writes fail after `BEGIN IMMEDIATE` (rolled back, so no corruption), but the error is opaque.

## Suggested fix

```python
ver = conn.execute("SELECT ver FROM col").fetchone()[0]
if ver < 18:
    raise RuntimeError(
        f"Unsupported collection schema {ver}; open the profile once in Anki >= 2.1.50 to upgrade it."
    )
```

Map to a dedicated error code (e.g. `COLLECTION_UNSUPPORTED`, exit 7) in the CLI layer.

Related nit: `_connect_write` (`:74-76`) runs `ROLLBACK` in the `except` even when `BEGIN IMMEDIATE` itself failed ("database is locked"), which raises "no transaction is active" and masks the real error. Guard with `if conn.in_transaction:`.

ubermenchh:
Closes #24

## Problem

`AnkiDirectReadStore.__init__` only checked that the file existed. The module assumes schema >= 18 (separate `decks`/`notetypes`/`fields`/`templates`/`deck_config` tables with protobuf blobs). A legacy schema-11 collection (Anki <= 2.1.27, or a profile never opened in a modern Anki) failed mid-command with `sqlite3.OperationalError: no such table: decks`.

## Change

- New `UnsupportedCollectionError` and `MIN_SUPPORTED_SCHEMA_VERSION = 18`. The store probes `col.ver` read-only (`mode=ro`, closed immediately) on construction and refuses `< 18` with a message naming the version, the path, and the fix (open once in Anki 2.1.50+).
- `create_backend_from_context` maps it to `BackendFactoryError`, so every command already emits `BACKEND_UNAVAILABLE` / exit 7 through the envelope — no per-command changes.
- Files without a readable `col.ver` (non-Anki SQLite, stripped-down fixtures) are left alone so the guard fires only on real legacy collections.

Verified end to end: `anki --backend direct --col legacy.anki2 decks` → JSON error, exit 7 (was a traceback).

## Tests

`test_anki_direct_schema_guard.py`: schema 11 refused with hint; 18 and 19 accepted; 17 boundary refused; conftest DDL accepted; no-col / no-ver / non-SQLite files pass through; the probe does not hold a lock. Factory test maps the error. 661 + 57 tui pass.

The `_connect_write` rollback nit mentioned in #24 landed in #47.


ubermenchh:
Review follow-ups in the second commit:

- **URI delimiters (important):** SQLite terminates a `file:` URI path at the first `#`/`?`, so a profile named `Deck#1` made the guard silently open a different file and skip. Switched to the plain-path + `PRAGMA query_only` style `_connect` uses (the file is known to exist, so rwc cannot create anything); this also removes the `mode=ro` asymmetry on WAL files with no `-shm`. Verified the new test fails against the URI implementation.
- `int(col.ver)` on a text value escaped as a traceback; now left alone like other unreadable versions.
- Tests: `#` and `?` parametrized separately (`?` skipped on win32), WAL-without-shm, empty `col` / NULL / text `ver`, and the left-alone cases now also assert a later query rejects the file.

`detect.py:280` has the same `file:{path}?mode=rw` pattern (pre-existing); leaving it for after #52 lands since that PR reworks `detect.py`.

PR #56, opened 2026-09-13
fix(direct): FSRS fidelity — learning step, weight upgrades, seeded fuzz, revlog kind, no lrt

ubermenchh:
Part of #22 — items 1–5 (scheduler fidelity). Item 6 landed in #48; item 7 (card generation from `reqs`) + the `create_notetype` `KIND_ALL` comment will be a separate PR so this one stays reviewable.

| # | Problem | Fix (rslib reference) |
|---|---|---|
| 1 | Learning step always reset to 0 → a card on its last step got the whole ladder again | `step = total_steps - (left % 1000)` (`left` packs `today_remaining*1000 + remaining`) |
| 2 | py-fsrs 6 rejects 17/19-weight sets → every answer in a deck last optimized on Anki 24.06–24.10 crashed | Port of fsrs-rs `check_and_fill_parameters_fsrs6`: 19 → `+[0, 0.5]`; 17 → rewrite w4/w5/w6 then `+[0,0,0,0.5]`; out-of-bounds → FSRS-6 defaults. Result payload reports `fsrs_params: fsrs6 \| fsrs5-upgraded \| fsrs4.5-upgraded \| default` |
| 3 | Preview and answer fuzzed independently → the interval shown ≠ the one written | Seed py-fsrs's RNG with `card_id + reps` per call (`get_fuzz_seed_for_id_and_reps`), restore caller state after |
| 4 | Review card answered Again logged as relearning (type 2) | Kind from the state *before* the answer: new/learn 0, review 1, relearn 2, and **3 (Filtered)** for a review answered early (`ReviewState::revlog_kind`, `days_late < 0`) |
| 5 | `cards.data["lrt"]` is not an Anki `CardData` key → dropped on Anki's next write | Removed; `last_review` comes from the latest real revlog row (types 0–3). Side effect: stored `s`/`d` is now used instead of being re-seeded from the revlog |

## Tests

`tests/integration/test_anki_direct_fsrs_fidelity.py` (17): step recovery incl. the issue's graduation repro; the three upgrade transforms against fsrs-rs's formulas; an FSRS-5 deck config answering successfully; out-of-bounds fallback; **preview == answer with the global RNG re-seeded in between** (simulating separate processes — the `getstate/setstate` alone made a naive version of this test pass, so the reseed is what actually kills the "no seeding" mutant); revlog kind for all six states incl. early review; last-review lookup ignoring manual entries; `lrt` scrubbed while custom data survives; stored memory state not re-seeded.

Mutants run and killed: step reset, no fuzz seeding, old Again→relearn rule, no early-review kind, 19 params passed through, `lrt` written.

Existing `answer_card` test updated: a review card answered Again now expects revlog type 1 (Review), with a comment citing rslib.

1584 + 68 tui pass. Real collection: preview Good = 137d/day 1616, answer wrote 137d/day 1616; revlog `(type 1, ivl 137)`; data keys `d, dr, pos, s` only.


ubermenchh:
Review follow-ups (second commit), each checked against rslib source rather than memory:

- **`lrt` — the review was right, the issue was wrong.** `rslib/src/storage/card/data.rs` has `#[serde(rename = "lrt")] last_review_time: Option<TimestampSecs>`, and `Card::days_since_last_review` prefers it. Restored: `lrt = now` is written on every answer; `_card_row_to_fsrs` reads it first and the revlog is the fallback. Tests flipped accordingly; #22 item 5 text is incorrect and I'll annotate the issue.
- **Early-review revlog kind — kept.** `rslib/src/scheduler/states/review.rs::revlog_kind` is `if self.days_late() < 0 { Filtered } else { Review }`, dispatched via `NormalState::Review(state) => state.revlog_kind()`. So rslib does log a review answered ahead of its due day as type 3; the test stays.
- **Clamp instead of default (suggestion taken):** upgraded weights are clamped to py-fsrs's `LOWER/UPPER_BOUNDS_PARAMETERS` (fsrs-rs `parameter_clipper`) so an FSRS-4.5 set whose upgraded `w4 = 2·w5 + w4 > 10` keeps the user's optimization; label becomes e.g. `fsrs4.5-upgraded-clamped`. Only unrecognized lengths / a corrupt `w5 ≤ -1/3` default.
- `math.log` domain guarded.

Out of diff, filed separately: empty `relearn_steps` gets a default 10 m step here while Anki keeps the card in Review on Again.

1587 + 68 tui pass.

PR #58, opened 2026-09-13
fix(direct): pass empty (re)learning steps through to the scheduler

### LINKED ISSUE #57: Empty relearning steps get a default 10m step; Anki keeps the card in Review
`anki_cli/db/anki_direct.py` `_build_scheduler`: `_to_timedeltas(cfg.relearn_steps, default=[10.0])` substitutes a 10-minute relearning step when the deck options have **no** relearning steps (a common FSRS setup). Both rslib and py-fsrs (`Scheduler.review_card`, `relearning_steps == []`) keep a review card in the Review state on Again with the FSRS-computed next interval; the CLI instead moves it to relearning (`type 3, queue 1`, due in 10 minutes).

Same for `learn_steps = []` (default `[1, 10]`): Anki graduates a new card straight to Review on the first answer.

Fix: pass an empty list through to `Scheduler(...)` when the config is empty (`_to_timedeltas` should only fall back to a default when the field is *unset*, which for repeated proto fields is indistinguishable from empty — so drop the defaults entirely and let py-fsrs use its own only if the deck_config row is missing). Test: `relearn_steps=[]`, review card Again → `type 2, queue 2`.

Found during the #56 review.

LouisDeconinck:
Closes #57

## Problem

A deck options row with blanked `learn_steps` / `relearn_steps` is a real configuration — Anki's preset editor allows it, and FSRS setups commonly drop relearning steps entirely. But `_to_timedeltas` treated empty as "unset" and substituted its own defaults (`[1, 10]` / `[10]` minutes), so:

- a review card rated **Again** was moved to relearning (`type 3, queue 1`, due in 10 minutes) instead of staying in Review with the FSRS-computed interval, which is what rslib and py-fsrs do when `relearning_steps == []`;
- a new card walked phantom learning steps instead of graduating straight to Review on the first answer, as Anki does with blanked learning steps.

## Changes

**`_build_scheduler`** — repeated proto fields cannot distinguish "unset" from "empty", so a present `deck_config` row now always supplies the step lists, empty included. Only when the row itself is missing are the step kwargs left out entirely, letting py-fsrs's built-in defaults apply. The returned step counts now reflect the scheduler's effective steps in both paths.

**`_to_timedeltas`** — the `default` parameter and the all-non-positive `[60s]` fallback are gone; the function converts what it is given and returns an empty list for an empty configuration.

## Tests

New `tests/integration/test_anki_direct_empty_steps.py` (6 tests):

- the prescribed regression: `relearn_steps=[]`, review card Again → `type 2, queue 2`, `due = today + ivl`;
- `learn_steps=[]` → new card graduates to Review on the first answer, on both Good and Again;
- scheduler level: empty step lists reach py-fsrs untouched (`()`, counts 0), while a missing `deck_config` row still gets the library defaults (`1m/10m`, `10m`);
- control case: populated `[1, 10]` / `[10]` steps still produce intraday learning/relearning (`type 1/3, queue 1`, epoch due, `left = 1001`).

## Not in this PR

Nothing — scoped to the step plumbing. Heads-up: this touches `_build_scheduler`, which #56 also refactors; whichever lands second will need a small rebase.


LouisDeconinck:
Merged latest `main` and resolved the conflict in `_build_scheduler`. Both changes are preserved: `_deck_config_for_deck` now also reports whether a config row exists (needed so a missing row still falls back to py-fsrs defaults while an existing row's empty step lists mean "no steps"), and the new step kwargs flow through upstream's `try/except ValueError` fallback to `FSRS6_DEFAULT_PARAMETERS`. Verified locally: `test_anki_direct_empty_steps`, `test_anki_direct_fsrs_fidelity`, `test_anki_direct_answer_card`, and `test_anki_direct_deck_config_and_add_notes` — 42 tests, all passing.

ubermenchh:
resolve the conflicts here

LouisDeconinck:
Conflicts resolved — merged latest `main`; the only overlap was the `typing` import line (kept `TYPE_CHECKING`/`TypedDict` for the step-kwargs typing). All 307 integration tests pass locally.

PR #59, opened 2026-09-13
fix(direct): generate cards from templates like Anki; derive reqs from templates

### LINKED ISSUE #22: FSRS fidelity gaps: step reset, FSRS-5 params, fuzz mismatch, reqs ignored, ORDER BY unit mix

ubermenchh:
Closes #22 (item 7 + the `create_notetype` comment; items 1–6 landed in #48/#56)

## Problem

`add_note` generated one card per template regardless of content: "Basic (and reversed card)" with an empty Back produced a blank reverse card that shows up in reviews and in Anki's *Empty Cards*. Separately, `create_notetype` wrote `reqs` as `KIND_ALL` over **every** field, which would have suppressed card creation unless every field was filled once `reqs` were honoured.

## What Anki actually does (verified in rslib)

- Generation does **not** read the `reqs` cache. `CardGenContext::new_cards_required_normal` parses each template's front and generates a card iff `renders_with_fields(nonempty_fields)` — a replacement of a non-empty field, or a `{{#x}}` block whose key is non-empty and whose body renders, or a `{{^x}}` block whose key is empty and whose body renders.
- `nonempty_fields` = note fields not matching `field_is_empty` (whitespace and empty `<br>`/`<div>` only — an `` is content) **plus** the special fields (`Card`, `CardFlag`, `Deck`, `Subdeck`, `Type`, `CardID`; `Tags` only when the note has tags; never `FrontSide`).
- A brand-new note always gets template 0 if nothing renders (`ensure_not_empty`).
- `reqs` is a legacy cache written for old clients via `ParsedTemplate::requirements` (any-single-field-suffices first, then jointly-required set, else none; negated conditionals ignored for the cache).

## Changes

- `core/template.py`: `field_is_empty`, `parse_template` (replacements with filters stripped, `{{#}}`/`{{^}}` sections, comment-wrapped directives), `template_renders_with_fields`, `template_requirements`, `SPECIAL_FIELDS`.
- `_template_ords_for_note`: implements the rule above; cloze ord capped at 499 like rslib.
- `_recompute_reqs(conn, ntid)` runs on `create_notetype`, `add_notetype_template`, `edit_notetype_template` (front changed), `add_notetype_field`, `remove_notetype_field` — replacing both the hand-written `KIND_ALL` and the manual renumbering from #40.

## Tests

- `tests/unit/test_template_cardgen.py` (20): rslib's own `requirements()` vectors incl. the comment-wrapped case and the `check_negated` asymmetry; `field_is_empty` table.
- Integration: one vs two cards on Basic (and reversed); `<br>`/whitespace-only vs ``; `ensure_not_empty`; conditional front and `Tags`/`Deck` special fields; `reqs` expectations corrected to Anki's (stock Cloze is `ANY [0]`).
- Mutants killed: generate-every-template, no ensure_not_empty, Tags-always-nonempty, br-counts-as-content, create_notetype-without-reqs.

1611 + 68 tui pass.

## Note on the real-collection comparison

Comparing our `requirements()` against the `reqs` stored in a real collection: every single-field template disagrees on the *kind* label (`any [0]` vs stored `all [0]`) — semantically identical for one field. rslib's algorithm (and its unit test `{{a}}{{b}}` → `Any`) checks "any single field suffices" first; the stored `all` values come from legacy clients / imports, and Anki only rewrites `reqs` on notetype save. Multi-field templates agree.

## Not in this PR

Anki also rewrites template text when a referenced field is removed (`update_templates_for_renamed_and_removed_fields`); we leave the `{{B}}` reference in place, so that template's `reqs` becomes `NONE` (which is what Anki would store *before* its rewrite). Worth its own small issue if wanted.


umangk-2410:
Review follow-ups in the second commit. The review couldn't fetch rslib, so I verified each point against `template.rs` / `notetype/mod.rs` / `cardgen.rs` first:

**Confirmed and fixed**
- `update_note` never generated cards → with `add_note` now withholding blank cards, filling a field later left the note short. Added `_generate_missing_cards` (rslib `generate_cards_for_existing_note`): only adds, never removes; new cards go to the deck of the note's existing cards. Result gains `generated_cards`.
- Field removal left `{{B}}` in templates → rslib rewrites templates first (`rename_and_remove_fields` + `add_missing_field_replacement`: unwrap sections, drop replacements, pad an empty front with the first remaining field, cloze on both sides). Ported; `reqs` recomputed from the rewritten templates.
- Parser strictness: rslib's `parse_inner` errors on unbalanced/mismatched sections → `template: None` → no card and `reqs` NONE. Now raises `TemplateParseError`, handled that way at both call sites. Contrary to the review, an unterminated `{{` is *not* an error in rslib (`handlebar_token` fails → `Text`); kept as text.
- `field_is_empty`: `[[:space:]]` is ASCII-only in the regex crate → `re.ASCII`. NBSP is content.
- `{{{Front}}}` → `classify_handle` does `trim_start_matches('{')` → handled.
- `has_tags` from the normalized tag string, in both `add_note` and `update_note`.
- Cloze: `_CLOZE_RE` (requires the closing `}}`); numbers above u16 aren't clozes.

**Refuted by source**
- Comment-wrapped directives: `next_token` tries a handlebar, then a comment, at each position, so `<!--{{^a}}-->` is a `Comment` token, ignored wholesale — my earlier regex that "un-commented" directives was wrong in the *other* direction. Comments are now opaque; `<!-- {{Back}} -->` no longer counts as a replacement.
- Key trimming: after `classify_handle` trims the whole handle, `rsplit(':')` does not re-trim, so `{{text: Front}}` has key `" Front"` in Anki. Matched.

The parser now preserves text/comments and round-trips via `nodes_to_string`, which is what made the template rewrite possible. Unit tests cover strictness, opaque comments, triple-brace, legacy `<% %>` syntax, and the rewrite cases; integration tests cover generation on edit (fields and tags), no removal on emptying, and the rewritten `{{A}}` front after removing `B`.

Mutants: no-generation-on-edit, no-template-rewrite, has_tags-from-raw-list — all killed. 1619 + 68 tui pass.

PR #61, opened 2026-09-13
fix(direct): reject duplicate and empty notes like Anki; scope the duplicate lookup by notetype

ubermenchh:
Part of #23 — the `allow_duplicate` / `mid` / single-csum half, plus the empty-first-field check that sits next to it in rslib. The csum HTML-strip is in #50; `sfld` stripping will follow once that merges so we don't both touch `_field_checksum`.

## Problem

`add_note` in `db/anki_direct.py` looked duplicates up by `csum` alone (across every notetype), wrote a free-text warning to stderr, and then inserted anyway. It also accepted a blank first field (`""`, `"   "`, `"<br>"`) and wrote a note plus a card for it. AnkiConnect refuses both:

```sh
anki note:add --deck Default --notetype Basic --Front hola --Back hello
anki note:add --deck Default --notetype Basic --Front hola --Back hello   # direct: succeeds; AnkiConnect: fails
anki note:add --deck Default --notetype Basic --Front "<br>" --Back hello # direct: succeeds; AnkiConnect: fails
```

The same command behaved differently per backend, and `note:bulk` could never surface a refusal because the store never raised.

## Change

Mirrors rslib `note_fields_check` (what AnkiConnect's `addNote` runs): **Empty** is checked first and is not lifted by `allow_duplicate`; then **Duplicate**, scoped by notetype.

- **`EmptyNoteError`** — raised when `field_is_empty(first_field)`, the same predicate card generation already uses, so `"<br>"` is empty for both. AnkiConnect: "cannot create note because it is empty".
- **`DuplicateNoteError`** — `allow_duplicate=False` is enforced. Carries `.notetype` and `.duplicate_ids` (all matches, ascending). Both share the `NoteRejectedError(ValueError)` base.
- Both are raised before `_allocate_row_id`, so nothing is written: no note, no card, `col.mod` untouched.
- **Lookup scoped by `mid`** via `_find_duplicate_note_ids(conn, *, notetype_id, csum)`: `csum = ? AND mid = ?`.
- **`csum` computed once** and reused for the `INSERT` (was computed twice from the same input).
- **CLI `note:add`:** catches `ValueError` alongside `AnkiConnectAPIError` / `LookupError` → `BACKEND_OPERATION_FAILED`, exit 1. For a duplicate, `details.duplicate_ids` lists the matches and the message gets `Pass --allow-duplicate to add it anyway.` appended — the store states the fact only; the remedy flag is CLI surface.
- **`add_notes` / `note:bulk`:** only *per-item* refusals (`NoteRejectedError`, `LookupError` for an unknown deck/notetype/field) become `null` ids, matching AnkiConnect's `addNotes` shape. Anything that would fail every item the same way — Anki Desktop holding the collection, a corrupt notetype config — now propagates and `note:bulk` fails the whole command with `BACKEND_OPERATION_FAILED`, instead of exit 0 with N `null`s that read like N duplicates.
- **`SKILL.md`:** documents both refusals, that `duplicate_ids` is direct-only (AnkiConnect reports the duplicate without ids), and what `null` in `note:bulk` does and does not mean.

## Tests

`tests/integration/test_anki_direct_note_mutators.py` — duplicate seeds go through the real `add_note` (`allow_duplicate=True`) rather than `_insert_note`, so the fixture csum is whatever the implementation computes; the tests hold regardless of how the helper populates `csum` (which #50 changes).

- duplicate in same notetype raises `DuplicateNoteError` with the right id; nothing persisted (note, card, `col.mod`)
- it is a `ValueError` and a `NoteRejectedError`
- `allow_duplicate=True` inserts the second note with the same `csum`
- two seeds → `duplicate_ids == [a, b]` ascending (pins no `LIMIT`, sort order)
- same first field under a different `mid` is **not** a duplicate
- empty first field is refused for `""`, whitespace, `<br>`, `<div></div>`, markup soup — with and without `allow_duplicate`; nothing persisted
- Empty is reported before Duplicate when an identical empty row already exists
- `_field_checksum` is called once per `add_note` (perf tripwire; behaviour is pinned by the csum-equality test)
- `add_notes`: duplicate / empty / unknown-deck items → `None`, good item lands; a write-guard refusal propagates and writes nothing
- the #59 `ensure_not_empty` test now reaches "nothing renderable" with a filled first field no template references

`tests/cli/test_note_commands.py`: `note:add` maps `DuplicateNoteError` → exit 1 / `BACKEND_OPERATION_FAILED` / `details.duplicate_ids` / remedy appended; `EmptyNoteError` → same code, no remedy, no ids; plain `ValueError` → same code, plain details; `note:bulk` with a propagated `RuntimeError` → one `BACKEND_OPERATION_FAILED`.

Mutation check: deleting the empty guard reds 7 tests; reverting `add_notes` to `except Exception` reds 1; dropping `AND mid = ?` reds 1.

## Verification

- `pytest`: 1659 passed; `pytest -m tui`: 68 passed
- `ruff check .` clean; `ty check anki_cli` clean
- CLI end-to-end on a tmp collection: add → id; add again → `BACKEND_OPERATION_FAILED` with `duplicate_ids` and the `--allow-duplicate` hint; `--Front "<br>"` → `Empty note…`; `--allow-duplicate` → id

## Not in this PR

- HTML-stripped `sfld` (item 5 in #23) — depends on `strip_html_preserving_media_filenames` from #50.
- Running the emptiness check on the HTML-stripped text (rslib does `strip_html(...).trim().is_empty()`); `field_is_empty` covers the common markup cases today. TODO(#23) in code.
- Anki additionally confirms a csum hit by comparing the stripped first field; with a 32-bit csum scoped per notetype the collision risk is negligible, so this relies on the csum match. TODO(#23) to add the confirm step after #50.
- `note:add` still surfaces a locked-collection `RuntimeError` as a traceback (pre-existing, every command). `note:bulk` catches it here only because this PR makes it propagate there; #50's `DirectWriteBlockedError` is the right hook for a repo-wide fix.

## Merge note vs #50

Both PRs touch the `csum` reuse hunk in `add_note` and the import block at the top of `test_anki_direct_note_mutators.py`. Whichever lands second: keep both sides.


ubermenchh:
Pushed 78b12be addressing the review on top of 5e1d7de:

- **`note:bulk --allow-duplicate`** threaded through `Protocol.add_notes(notes, *, allow_duplicate=False)`, `DirectBackend`, `AnkiConnectBackend` (per-note `options.allowDuplicate`, since `addNotes` has no top-level option) and the store. Re-running an import file now has a way in again; empty refusals are still not lifted.
- **README** documents the refusal and the flag; SKILL.md notes same-batch repeats are duplicates.
- `note:add` uses `isinstance(exc, DuplicateNoteError)` instead of duck-typing `duplicate_ids`.
- `DuplicateNoteError` message caps at 10 ids + "and N more"; `.duplicate_ids` keeps the full list.
- New tests: same-batch repeat refused, `allow_duplicate` inserts repeats / doesn't lift empty, missing field precedes duplicate, message truncation, delegation and AnkiConnect payload with the flag, CLI flag passthrough.

Still deferred to after #50 (both need the stripper): confirming a csum hit against the stripped first-field text, and running the emptiness check on stripped text. Both are marked `TODO(#23)` in `anki_direct.py`.

`pytest`: 1667 passed; `ruff check .` and `ty check anki_cli` clean.

PR #63, opened 2026-09-14
fix(search): align filter semantics with Anki — is:due, deck:/tag: hierarchy, added:, unknown prefixes

ubermenchh:
Part of #30 — the **search-semantics** slice (`core/search.py`, `_deck_filter`, AC `get_due_counts`). Return-shape normalization / Protocol additions / `_store` probes are left for after #52, since they need `cli/commands/review.py` and `cards.py`.

## Problem

The direct backend's query compiler diverged from rslib's `sqlwriter` in several places, so the same string meant different things per backend (issue #30, "Query semantics differ"):

| Query | Direct (before) | Anki / AnkiConnect |
|---|---|---|
| `is:due` | includes every new card (`queue = 0`) | learning/review cards past due only |
| `deck:Lang` | `Lang` only | `Lang` **and** `Lang::Spanish`, incl. cards visiting a filtered deck (`odid`) |
| `review --deck Japanese` | exact name | includes `Japanese::Core` |
| `tag:verb` | `verb` only | `verb` and `verb::irregular` |
| `added:1` | `notes.mod` within 24h (any edit counts) | card created since last rollover |
| `card:1`, `rated:1`, `mid:…` | full-text search for the literal string | real filters |
| AC `get_due_counts` new | `is:due is:new` → always 0 | `is:new` |

## Change

**`core/search.py`**
- `is:due` drops `queue = 0` (rslib `StateKind::Due`).
- `deck:` compiles to `c.did IN (decks WHERE name LIKE p OR name LIKE p||'::%') OR (c.odid != 0 AND c.odid IN (...))` (rslib `write_deck`). The `LEFT JOIN decks` is gone; the note-level form wraps the same clause in `EXISTS`.
- `tag:` matches `% x %` or `% x::%` (rslib `write_tag`); `tag:none` → `n.tags = ''`.
- `added:N` → `c.id > (next_day_at - N*86400) * 1000` (rslib `write_added`). The compile entry points now take a `SearchContext(now_sec, due_day_index, next_day_at)` instead of two loose ints; `anki_direct` builds it from the existing `_timing`.
- Unknown prefixes raise `SearchParseError("Unsupported filter 'card:'. Supported: … To search for a literal colon, escape it as '\:'.")` with the token position → `INVALID_INPUT` at the CLI. Anki treats unknown prefixes as field searches; this CLI does not implement those, and the old full-text fallback was silently wrong for real Anki filters.
- `note:` accepted as Anki's spelling of `notetype:`.
- Literal colon: the tokenizer records the index of the first **unescaped** colon on the `Token`, so `a\:b` is a text search for `a:b` while `"deck:x"` is still a deck search (matches Anki: quotes don't protect the colon, only `\:` does).

**`db/anki_direct.py`**
- `_deck_filter` (used by `get_due_counts` / `get_next_due_card`, i.e. `review --deck` and `deck` stats) now selects the deck **and** its `::` children, `COLLATE NOCASE` (deck names are unique case-insensitively in Anki), with LIKE metacharacters escaped. It still keys on `did` only — Anki's deck list counts a card under the deck it currently sits in.
- `find_card_ids` / `find_note_ids` go through a new `_search_context()`.

**`backends/ankiconnect.py`** — `get_due_counts` uses `is:new` for the new count.

**Docs** — README and SKILL.md search sections describe the hierarchy, `added:`, `tag:none`, `note:`, `\:`, and that unsupported prefixes are rejected.

## Behaviour changes to be aware of

- `is:due` results and any `--query is:due` action (`card:suspend --query is:due` etc.) no longer touch new cards. Direct `get_due_counts` was already counting `queue = 0` separately, so review totals don't change.
- `deck:`/`tag:`/`--deck` now include children. A script that relied on `deck:Parent` *excluding* subdecks will see more cards.
- Queries with an unknown `prefix:` now fail with exit 2 instead of returning (wrong) results.
- `--deck` on `review`/`decks`/`deck` covers subdecks, so parent rows in `decks` include child counts.
- `is:new` / `is:review` go by card type (suspended new cards are `is:new`; relearning cards are `is:review`), matching AnkiConnect.

## Tests

`tests/integration/test_anki_direct_find_queries.py` (fixture gains `odid`): parent/child/grandchild vs `Language` sibling, case-insensitivity, `-deck:`, filtered-deck `odid`, tag children + glob + `tag:none`, `note:` alias, `added:` with three cards straddling two rollovers while every note has `mod = now` (the old implementation returned all three for `added:1`), six unknown prefixes with position `0`, escaped/quoted colons.
`tests/integration/test_anki_direct_due.py`: the old `deck_filter_is_exact_name_match` test asserted the bug; replaced with children-included, `DeckAB` not a child, `decka` case-insensitive, `100%` escaping.
`tests/backends/test_ankiconnect.py`: pins the three due-count queries.

`pytest`: 1675 passed · `ruff check .` · `ty check anki_cli` clean.

## Not in this slice

- Learn-ahead limit for intraday `is:due` (Anki adds `learn_ahead_secs`, default 20 min).
- Field searches (`front:dog`), `deck:current` (rejected at parse time), `rated:`, `prop:due` day-relative semantics, Unicode case folding.
- Everything under "Return shapes differ" / "Methods missing from the Protocol" in #30.


umangk-2410:
Pushed 86c876d addressing the review:

- **`is:new` / `is:review` by card `type`** (rslib `write_state`); `is:due` includes queue 4. Both backends now agree. The "not in this slice" note in the description is obsolete — done here since README claims parity.
- **`tag:*`** → every note; tag with a space → nothing; `tag:none` uses `TRIM`.
- **`deck:*`**, **`deck:filtered`** (`odid != 0`); `deck:current` rejected at parse time.
- **`added:0`** clamps to 1, negatives are a parse error (rslib `parse_added`).
- `_escape_like` alias dropped.
- **Tests that passed for the wrong reason fixed**: `added:` now has a card created 23h ago but before the last rollover (kills a `now_sec` cutoff mutant) and one exactly on the cutoff (pins `>`); the LIKE-escape test has `100X::Sub`/`100Y::Sub`/`100_` decoys so dropping `escape_like` fails. Tag child anchor pinned with `adverb::irregular`.
- New `tests/unit/test_search_parse.py`: separator/escape logic at parse level, including the leading-whitespace-in-quotes shift and non-zero error positions.
- README/SKILL: `--deck` on `review`/`decks`/`deck` covers subdecks (parent rows in `decks` include child counts — 4th behaviour change); `deck:filtered`, `tag:*`, ASCII-only case folding.

`pytest`: 1705 passed; ruff / ty clean.

Still deferred: learn-ahead window on intraday `is:due`, Unicode case folding, field searches, `rated:`.

Issue #29, opened 2026-09-12
Collapse cli/commands boilerplate into an @anki_command decorator; dedupe card rendering


ubermenchh:
## Problem

`anki_cli/cli/commands/*.py` is mostly copy-pasted scaffolding. Counts across the directory:

| Pattern | Occurrences |
|---|---|
| `obj: dict[str, Any] = ctx.obj or {}` | 54 |
| `formatter = formatter_from_ctx(ctx)` | 65 |
| `with backend_session_from_context(obj)` | 47 |
| `except (BackendNotImplementedError, BackendFactoryError, NotImplementedError)` | 47 |
| `raise click.exceptions.Exit(` | 93 |
| `code="BACKEND_OPERATION_FAILED"` | 28 |
| `"Provide --id or --query."` | 9 |
| `_emit_backend_unavailable` copy-pasted | 7 files (`cards.py:20`, `note.py:22`, `tag.py:18`, `search.py:18`, `notetype.py:17`, `review.py:21`, `deck.py:17` as `_emit_backend_error`) |
| `_emit_invalid_query` copy-pasted | 4 files |

Card rendering (`_extract_note_id` / `_extract_ord` / `_pick_template` + the cloze/normal `render_template` block) exists **4×**: `cards.py:87-215`, `review.py:55-172`, `tui/review_app.py:48-78, 770-828`, `tui/repl.py:513-599`.

Catching Python's builtin `NotImplementedError` in 47 places (to handle the one AnkiConnect method that isn't implemented, `get_revlog` at `ankiconnect.py:631`) will also swallow genuine unimplemented-code bugs as "backend unavailable".

## Suggested fix

```python
@anki_command("card:suspend", help="...")
@click.option("--id", ...)
def card_suspend(backend: AnkiBackend, *, id: int | None, query: str | None) -> JSONValue:
    ids = resolve_ids(backend, id, query)
    backend.suspend_cards(ids)
    return {"suspended": ids}
```

where `anki_command`:
- reads `ctx.obj`, builds the formatter, opens the backend session;
- calls the function and emits `ok` envelope with the return value;
- maps exceptions through one table (see #27).

Plus:
- A shared `id_or_query` option group.
- `core/render.py::render_card(backend, card_id) -> RenderedCard` used by CLI and TUI.
[...]

Issue #28, opened 2026-09-12
Docs vs CLI drift: 'cards' is a TUI, note:bulk schema, exit codes, inconsistent flag names


ubermenchh:
## Problem

Docs and the CLI disagree in several places that agents will hit immediately.

### 1. `cards` is a TUI, docs say it's the JSON query command
`cli/commands/browse.py:16-20, 49` registers `cards` as "Browse cards interactively (TUI)". `README.md:51, 124-127, 135` and `SKILL.md:33, 88, 278` tell agents to run `anki --format json cards --query ...`. In a non-TTY this launches Textual instead of emitting JSON. The list variant is hidden as `cards:ids` (`cards.py:63`) and returns ids only; `search` (`search.py:56`) is the only full-detail lister.

Fix: rename the TUI to `browse` (or `cards:browse`); make `cards` return detail rows (what `search` does today) and fold `search` into it.

### 2. `note:bulk` schema
`note.py:345-352` requires `[{"fields": {...}, "tags": [...]}]`; `SKILL.md:128, 294` documents flat `[{"Front": "Q1", "Back": "A1"}]`. Every documented bulk import fails with `INVALID_INPUT`.

Fix: accept both (if no `fields` key, treat the object minus `tags` as fields) and document one.

### 3. Exit codes / error codes
- Exit code **3** (`detect.py:57`, "Direct backend forced, but no collection DB found") is not in `README.md:210-216` or `SKILL.md:71-77`.
- `INVALID_CONFIG` (`app.py:116`, `config.py:93`) is not in `SKILL.md:81`.
- `LookupError` → exit 4 (`ENTITY_NOT_FOUND`) in `card`/`note`/`deck` but exit 1 (`BACKEND_OPERATION_FAILED`) in `card:move`, `tag:rename`, `note:fields`, `notetype:field:*`.
- `UNDO_EMPTY` exits 2 (`review.py:340`) but is semantically "not found".

Fix: `ExitCode` and `ErrorCode` enums in `models/output.py`; generate the docs table from them.

### 4. Inconsistent flag names for the same concept
- Deck identity: `deck --deck`, `deck:delete --deck`, `deck:config --deck`, but `deck:create --name`.
- Notetype: `notetype --notetype/--name`, `notetype:create --name`, `notetype:field:add --notetype`.
- `note:fields --fields` (comma-separated) vs `notetype:create --field` (repeatable).
- `--tag` is singular but accepts a space/comma list (`tag.py:66`).

[...]

PR #65, opened 2026-09-17
fix(cli): JSON envelope for every exit; accept global options after the subcommand

### LINKED ISSUE #26: --yes / --format only accepted before the subcommand; all docs put them after
## Problem

`--yes` (and `--format`, `--backend`, `--col`) are defined on the root group only (`anki_cli/cli/app.py:64`). Every documented example puts them **after** the subcommand:

- `README.md:156, 169, 238`
- `SKILL.md:140-141, 306`

```sh
$ anki --format json note:delete --id 1 --yes
Usage: anki note:delete [OPTIONS]
Error: No such option: --yes
```

Exit 2, **plain text** despite `--format json` (see #27). The error hints in `note.py:280` / `deck.py:263` ("Run with --yes before the command") acknowledge the problem instead of fixing it.

This is the single most likely failure an agent following SKILL.md will hit.

## Suggested fix

Either:

1. **Per-command `--yes`** via a shared decorator applied to every destructive command (`note:delete`, `deck:delete`, `card:reset`, `notetype:*:remove`, `tag:remove`, …), reading `ctx.obj["yes"] or yes`; or
2. In `NamespaceGroup.parse_args`, **hoist known global flags** out of the subcommand tail before dispatch.

Option 1 is more conventional and discoverable in `-h`. Keep the group-level flag working too.

Fix the docs either way, and add a `CliRunner` test that `note:delete --id 1 --yes` (flag after command) succeeds.

### LINKED ISSUE #27: Central exception -> JSON envelope mapper (tracebacks and Click usage errors bypass --format json)
## Problem

The JSON error envelope is not guaranteed. Several failure classes bypass it and emit plain text or Python tracebacks even with `--format json`:

1. **`AnkiConnectUnavailableError`** (`backends/ankiconnect.py:102-114`, raised on timeout or connection drop *mid-command*) is caught by **zero** command handlers. Only `AnkiConnectAPIError` (and occasionally `AnkiConnectProtocolError`, e.g. `review.py:229`) are. Result: traceback, exit 1.
2. **`sqlite3.OperationalError`** ("database is locked") is only caught in `detect.py:284`, never in commands.
3. **Click `UsageError`s** (unknown option, missing `--id`, unknown command) print Click's plain-text usage to stderr with exit 2, ignoring `--format json` (verified with `--yes` in the wrong position).
4. `KeyboardInterrupt` → traceback.

## Suggested fix

A single exception → envelope mapper, applied once:

```python
ERROR_MAP: dict[type[BaseException], tuple[str, int]] = {
    SearchParseError: ("INVALID_INPUT", 2),
    click.UsageError: ("INVALID_INPUT", 2),
    BackendFactoryError: ("BACKEND_UNAVAILABLE", 7),
    BackendNotImplementedError: ("BACKEND_UNAVAILABLE", 7),
    AnkiConnectUnavailableError: ("BACKEND_UNAVAILABLE", 7),
    sqlite3.OperationalError: ("COLLECTION_LOCKED", 7),
    LookupError: ("ENTITY_NOT_FOUND", 4),
    AnkiConnectError: ("BACKEND_OPERATION_FAILED", 1),
    ValueError: ("BACKEND_OPERATION_FAILED", 1),
}
```

- Wrap `main()` in a custom entry that invokes the group with `standalone_mode=False`, catches `click.UsageError` and `Exception`, and emits through the formatter when `--format json` (parse `--format` from argv early, before Click, so usage errors can honor it).
- This also enables the `@anki_command` decorator refactor (see #29), since the ~47 identical `except (...)` blocks become redundant.

Add `CliRunner` tests: unknown option with `--format json` → JSON envelope, exit 2; backend raising `AnkiConnectUnavailableError` mid-command → `BACKEND_UNAVAILABLE`, exit 7.

ubermenchh:
Closes #27. Closes #26.

## Problem

`--format json` did not hold for the failures an agent following SKILL.md hits first:

| Input | Before | Now |
|---|---|---|
| `anki --format json note:delete --id 1 --yes` | Click plain text `No such option: --yes`, exit 2 | runs (`--yes` hoisted) |
| `anki --format json nope:cmd` | plain text | `INVALID_INPUT` envelope, exit 2 |
| `anki --format json cards:ids --bogus` | plain text | `INVALID_INPUT` envelope with `details.usage`, exit 2 |
| `AnkiConnectUnavailableError` mid-command | traceback, exit 1 | `BACKEND_UNAVAILABLE`, exit 7 |
| `sqlite3.OperationalError: database is locked` / `DirectWriteBlockedError` | traceback | `COLLECTION_LOCKED`, exit 7 |
| Ctrl-C | traceback | `INTERRUPTED`, exit 130 |
| any bug | traceback | `INTERNAL_ERROR`, exit 1, `details.exception`; `ANKI_CLI_DEBUG=1` appends the traceback |

## Change

**`cli/app.py`**
- `NamespaceGroup.main()` runs Click with `standalone_mode=False` and renders anything that escapes through `OutputFormatter`. Envelopes commands already emitted arrive as `Exit(n)` and pass through untouched (no double reporting). `meta.command` is the subcommand when known, `bootstrap` otherwise.
- When the error fires before the group callback built `ctx.obj` (unknown command, bad group option), `--format`/`--no-color` are peeked from argv so the envelope still honours them.
- `NamespaceGroup.parse_args` calls `hoist_group_options` (#26): group option spellings after the subcommand move before it with their value tokens, unless the subcommand defines the same spelling; `-h`/`--help` and everything after `--` stay. The REPL uses `cmd.make_context` directly and is unaffected.

**`cli/errors.py`** (new) — `ERROR_MAP: type → (ErrorCode, ExitCode)`; `classify(exc)` walks the MRO so a subclass entry always beats its base; special cases for `DetectionError.exit_code`, sqlite "locked/busy", `click.UsageError` (`details.usage`).

**`models/output.py`** — `ExitCode` / `ErrorCode` enums + `EXIT_CODE_MEANINGS`. New codes: `COLLECTION_LOCKED`, `INTERRUPTED`, `INTERNAL_ERROR`; new exit 130.

**Docs** — README/SKILL exit- and error-code tables regenerated from the enums; new "Global Options Placement" section; `tests/unit/test_docs_sync.py` fails if the tables drift. The three `"Run with --yes before the command"` hints are now `"Re-run with --yes."`.

## Tests

`tests/cli/test_app_error_envelope.py` (28): unknown option/command/group value → envelope; `--help`/`-h`/`--version` still plain, exit 0; 9 escaping exception types → expected code/exit; internal error names the type; `ANKI_CLI_DEBUG` traceback; Ctrl-C → 130; a command's own envelope is not wrapped twice; `--yes`/`--format`/`--format=json` in six positions; a subcommand value that looks like a group flag is not hoisted; subcommand-owned spelling wins; nothing after `--` moves.
`tests/unit/test_params.py` (+16 hoist cases), `tests/unit/test_cli_errors.py` (classify table, MRO precedence, usage details, argv peek).

All 1823 pre-existing tests pass unchanged, including every `CliRunner.invoke(app_mod.main, …)` bootstrap test. `pytest`: 1893 passed · `ruff` · `ty` clean.

## Not in this PR

- Collapsing the ~47 per-command `except` blocks into `@anki_command` (#29). They stay as the first line; this is the net beneath them.
- Rest of #28 (command renames, `note:bulk` schema, flag-name consistency).


ubermenchh:
Pushed the review follow-ups:

- **Fixed the real bug**: non-Click exceptions now render through the resolved runtime (`make_context` records the root ctx; `_emit_escaped_error` falls back to it), so config/env `json` and `meta.backend` are honoured. The escaping-exception tests now run with the format coming from config only and assert `meta.backend == "direct"` — both failed before the fix.
- Dead code gone: `except Exit`, `standalone_mode=False` passthrough, four unreachable `ERROR_MAP` rows, module-global `main` reference.
- Lock markers use SQLite's real wording; `"busy"` false positive pinned by a test.
- `peek_output_options` / `command_name_from_argv` live in `params.py`, arity-aware and case-insensitive.
- `--version` stays with the subcommand like `--help`.
- `ErrorInfo.code: ErrorCode` — existing literals validate, unlisted codes fail at emit time.
- REPL honours trailing `--yes`/`--format`/`--copy`/`--no-color` per line, matching the docs.
- Test gaps closed per the review (classify rows, ctx-less `UsageError`, `sys.exit` passthrough, subcommand `--help`, docs_sync section slice, peek arity/case).

`pytest`: 1926 passed (+68 TUI); ruff / ty clean.

ubermenchh:
Round-two follow-ups pushed:

- **Hoister no longer lets a typo'd subcommand option swallow a trailing group option** — `probe --bogus --format json` renders the usage error as JSON again (reproduced before, pinned by test).
- **Dangling valued option** (`note:delete --id 1 --format`) is a clean `'--format' requires an argument` usage error in both CLI and REPL instead of reading the command name as the value.
- **REPL** passes its global arities to `preprocess_argv`, so `deck:delete --yes deck=Old` works (reproduced before, pinned). Trailing `--format` is validated like `set format`; alias commands tested.
- `command_name_from_argv` is arity-aware and judges the first positional: `nope probe` → `bootstrap`, `--col probe cards:ids` → `cards:ids`. `_emit_escaped_error` tries `ctx.invoked_subcommand` first.
- `KeyError`/`IndexError` → `INTERNAL_ERROR` with `details.exception`; bare `LookupError` stays `ENTITY_NOT_FOUND`.
- `_root_ctx` reset pinned by a test that fails root `parse_args` after a seeded run.

`pytest`: 1939 passed (+72 TUI); ruff / ty clean.

ubermenchh:
Round three found one remaining REPL gap: a mistyped *leading* option (`--yess note:delete --id 1`) was silently dropped and the command ran without it. Fixed in the latest commit — it now errors `No such option: --yess` and does not run — with tests for both the typo and the session-fixed `--backend` shape. No new regressions from the round-two fixes; the correctness pass verified the typo-swallow rule against `note:add`'s dynamic `--Field` options, the dangling-option off-by-one, `invoked_subcommand` on the no-args path, and that no app code raises `KeyError`/`IndexError` for not-found (so the `INTERNAL_ERROR` flip only affects genuine bugs).

`pytest`: 1939 passed (+74 TUI); ruff / ty clean.

PR #66, opened 2026-09-18
fix(cli): make the docs true — cards is JSON, browse is the TUI, note:bulk takes the documented shape

### LINKED ISSUE #28: Docs vs CLI drift: 'cards' is a TUI, note:bulk schema, exit codes, inconsistent flag names

ubermenchh:
Closes #28. (Item 3, exit/error code tables, landed in #65.)

Every item was the first thing an agent following SKILL.md would hit.

| # | Before | Now |
|---|---|---|
| 1 | `anki --format json cards --query …` launched the Textual TUI into a pipe; the JSON lister was `search`, ids-only was `cards:ids` | `cards` = full-detail JSON lister; `search` hidden alias; TUI is **`browse`** (REPL alias `b`) |
| 2 | `note:bulk` rejected the documented flat `[{"Front": "Q1", "Back": "A1"}]` | accepts flat **and** `{"fields": …, "tags": …}`; both normalize identically |
| 4 | `deck:create --name`, `notetype:create --name`, `tag:add --tag` only, `note:fields --fields a,b` | canonical `--deck` / `--notetype` / `--tags` / repeatable `--field`; old spellings kept as aliases; docs use the canonical form |
| 5 | `decks` data differed by `--format` (`total` vs `total_due`, pre-indented `name`) | one shape; table renderer indents `A::B` and hides `level` |
| 6 | `review:answer` wrote a free-text warning to stderr | `meta.warnings`; SKILL.md now promises stderr is empty on success |
| idea | — | **`anki commands --format json`**: every command, option (spellings/required/flag/type/help), global options, exit + error codes, from the Click registry |

## Drift guard

`tests/unit/test_docs_sync.py` now parses every fenced `anki …` line in README.md and SKILL.md and asserts the command exists and each `--option` is defined on it (dynamic `--Field` options on `note:add`/`note:edit` excepted), plus that `browse` never appears with `--format json`. Both docs pass clean after the edits — a future rename or flag change that isn't reflected in the docs fails CI.

## Behaviour changes

- `anki cards` no longer opens the TUI. Use `anki browse`. (`c` in the REPL still means `cards`; `b` is `browse`.)
- `decks --format table` rows carry `total_due` (was `total`) and the hierarchy indentation is presentation-only.
- `review:answer` no longer writes to stderr on an undo-store failure.

## Tests

`pytest`: 1954 passed + 74 TUI · `ruff` · `ty` clean. New: `cards`/`search` parity and hidden flag, `cards` without `--query`, `browse` is the TUI; `note:bulk` flat + nested + fieldless rejection; `note:fields --field`; `deck:create --deck|--name`, `notetype:create --notetype|--name`, `tag:add --tag|--tags`; `decks` shape identical across formats + renderer indentation leaves data untouched; `review:answer` warning in `meta.warnings` with empty stderr; `commands` introspection; docs-sync invocation check.


ubermenchh:
Review follow-ups pushed:

- **REPL regression fixed**: renaming the TUI left `c` outside the deck-context block, so `use Japanese` + `c` listed the whole collection and `c is:due` was a usage error. New `_prepare_line` (unit-tested) rewrites a bare positional to `--query` and folds `use <deck>` into `--query` for query commands / `--deck` for others; explicit `deck:` or `--deck` wins.
- **`note:bulk` flat form** refuses `deck`/`deckName`/`notetype`/`modelName`/`options` keys (they were being silently dropped, note landing in the `--deck` deck with `ok: true`); `"tags": null` is no tags instead of a literal `None` tag.
- **`cards --limit`** (default 1000, `0` = all) with `total` and a `meta.warnings` entry when truncated — the unbounded N+1 the docs now advertise.
- Table leaf uses the same rule as `decks`; `commands` exercised through `main()`.

`pytest`: 1962 passed + 89 TUI; ruff / ty clean.

PR #67, opened 2026-09-18
refactor(cli): @anki_command owns the envelope, session and BACKEND_UNAVAILABLE; one render_card

### LINKED ISSUE #29: Collapse cli/commands boilerplate into an @anki_command decorator; dedupe card rendering

ubermenchh:
Closes #29.

## What

`cli/commands/*.py` was 3032 lines, roughly half of it the same scaffolding in every handler. New `cli/command.py`:

```python
@anki_command("deck", errors={LookupError: ("ENTITY_NOT_FOUND", 4)})
@click.option("--deck", "deck_name", required=True)
def deck_cmd(cmd: CommandContext, deck_name: str) -> JSONValue:
    name = deck_name.strip()
    if not name:
        raise cmd.invalid("Deck name cannot be empty.")
    with cmd.errors(details={"deck": name}):
        return cmd.backend.get_deck(name)
```

`CommandContext` gives a lazy `backend` (session opened on first use, closed on exit), `invalid()` / `fail()` / `require_yes()` / `require_ids()` / `query_errors()`, and an `errors()` block that maps the command's table — most specific class first via MRO — with `details` / `message` callables that may read the exception. Return value → success `data`; `None` → nothing (TUI launchers). `BACKEND_UNAVAILABLE` for `(BackendFactoryError, NotImplementedError)` is built in. Unmapped exceptions fall through to the #65 entry-point mapper. `id_or_query(entity)` is the shared `--id` / `--query` pair.

`core/render.py::render_card` replaces the four copies of card rendering (`cards`, `review`, `ReviewApp`, REPL inline).

| | before | after |
|---|---|---|
| `cli/commands/*.py` lines | 3032 | 1503 |
| `except` clauses | 191 | 14 |
| `raise click.exceptions.Exit` | 95 | 1 |
| `_emit_backend_unavailable` copies | 7 | 0 |
| card-render implementations | 4 | 1 |

## Why the exception tables stay per command

The inventory before writing this showed the mapping is **not** uniform: `LookupError` is exit 4 in 7 commands and exit 1 in ~20; `ValueError` is 2 / 4 / 1 depending on the command; `AnkiConnectAPIError` is `INVALID_INPUT` on the listers and `BACKEND_OPERATION_FAILED` on mutations; ~60 test assertions pin exact `details` dicts. The issue asks for "no behavior change", so a single global table was off the table. Each command declares its own; the decorator removes the scaffolding around it. Normalising the codes is #30/#28-territory for a later PR where the change is the point.

## No behaviour change — verified

All **1962 pre-existing tests pass unmodified**, including every `CliRunner` test that monkeypatches `backend_session_from_context` / `formatter_from_ctx` on a command module (the decorator resolves both through the defining module's globals) and the deck test whose `CaptureFormatter.emit_success` has no `warnings` kwarg (the decorator only passes it when non-empty). `status` and `config:*` are left as-is: no session, bespoke `ctx.obj` reads, nothing to gain.

Two TUI tests needed a `(?i)` on an error-message regex because `render_card`'s messages are capitalised (`"Card has no note id."`) where `ReviewApp`'s copy was lowercase.

## Tests

New `tests/cli/test_anki_command.py` (18) pins the decorator in isolation: success / `None`, lazy session, `warnings` only when present, built-in BU for both types, MRO precedence, unmapped escape with session cleanup and no envelope, `details`/`message` callables, `require_yes`, `require_ids` + `id_or_query`, `query_errors`' `ankiconnect=` switch, `CommandExit` not re-wrapped, registration. README gains an "Adding a command" section (docs-sync passes).

`pytest`: 1980 passed + 89 TUI · `ruff` · `ty` clean.

## Follow-ups (not here)

- Implement `get_revlog` on AnkiConnect (`getReviewsOfCards`) and drop `NotImplementedError` from the BU tuple — it's now in one place, so that's a two-line change once the backend method exists.
- `note:bulk`'s `RuntimeError` row (pre-existing) still swallows `AnkiConnectUnavailableError` as exit 1; preserved verbatim, flagged for #30.


ubermenchh:
Review found one real regression, now fixed and pinned:

- **`note:bulk` returned exit 1 / `BACKEND_OPERATION_FAILED` on a session-open failure** (was 7 / `BACKEND_UNAVAILABLE`). Its `RuntimeError` row captured `BackendFactoryError`, a `RuntimeError` subclass, before the built-in check ran. `errors()` now decides backend-unavailable first unless the exact type is named. Reproduced before fixing; no existing test covered a factory failure on that command.
- New `test_backend_unavailable_parity.py` drives **every** registered backend-using command through a factory failure and asserts the 7 / BU / `{"backend"}` contract — the oracle that would have caught it.
- `card:suspend`/`unsuspend`: `find_cards` back inside the details block so an AnkiConnect `--query` failure keeps `{"id","query"}`.
- Dropped the `__globals__` lookup seam and the `# noqa: F401` import ritual it forced on every command module; tests patch `anki_cli.backends.factory` / `anki_cli.cli.formatter` directly (as `tests/tui` already did). `errors()` shed its unused knobs; `note:add` handles `DuplicateNoteError` explicitly.

`pytest`: 2031 passed + 89 TUI; ruff / ty clean.

PR #68, opened 2026-09-18
refactor(backends): scheduler introspection on the Protocol; retire backend.name / _store probes

ubermenchh:
Part of #30 — item 2 (methods missing from the Protocol) and the `answer_card` documentation note. Items 3/4 (search semantics, AC due counts) landed in #63. Item 1 (return-shape normalisation) will be its own PR: it changes the JSON agents see from the AnkiConnect backend and needs a documented migration.

## Problem

`get_next_due_card`, `preview_ratings`, `snapshot_card_state`, `restore_card_state` existed only on `DirectBackend._store`. Every consumer — `review:next/preview/undo/answer/start`, `ReviewApp`, the REPL inline reviewer — probed `backend.name == "direct" and hasattr(backend, "_store")` (14 sites) and reached into a private attribute.

## Change

- **`AnkiBackend` Protocol** gains the four methods under "Scheduler introspection" plus `supports_scheduler_introspection: bool`. `answer_card`'s docstring spells out that it means different things per backend.
- **`DirectBackend`** delegates them to the store like every other method; `supports_scheduler_introspection = True`.
- **`AnkiConnectBackend`** raises new `BackendUnsupportedError(NotImplementedError)` with `.operation`, `.backend` and a `--backend direct` hint; flag `False`. Being a `NotImplementedError` it lands in the CLI's existing `BACKEND_UNAVAILABLE`/7 path unchanged. The AC "no current GUI card" error now tells the user what to do.
- **Consumers** branch on the capability flag and call the method on the backend. `_store` is referenced nowhere outside `backends/`.
- **SKILL.md** documents which review commands are direct-only and the `answer_card` difference.

## Behaviour

Same codes and exits. `review:preview`/`review:undo` on AnkiConnect still exit 7 / `BACKEND_UNAVAILABLE`; the **message text changes** to `<command> is not supported by the ankiconnect backend. Use --backend direct.` (was `<command> is supported only for direct backend.`).

## Tests

- `tests/backends/test_protocol_parity.py`: both backends implement every Protocol method and declare the flag; AC raises `BackendUnsupportedError` for each with operation/backend/hint; **a source scan asserts no `_store` / `name == "direct"` probe exists in `cli/`, `tui/`, `core/`** so it can't creep back.
- `test_direct_backend.py`: delegation rows for the four methods.
- The 11 CLI + 3 TUI fakes that modelled the old contract (`name="direct"` + a `_store` holding the methods) now model the new one (methods on the backend + the flag) — the tests themselves are unchanged.

`pytest`: 2044 passed + 89 TUI · `ruff` · `ty` clean.


ubermenchh:
Review: no correctness findings; all 14 hunts clean (flag ⇔ old predicate for both real backends; BU routing preserved; REPL/ReviewApp/CLI gates equivalent; SKILL ordering matches scheduler.py). One nit fixed: `review:preview` named the Protocol method in its error while `review:undo` named the command, and the two hints differed. Both now read `<command> is not supported by the ankiconnect backend. Use --backend direct.` — pinned. PR description amended: codes/exits unchanged, message text updated.

Flagged for a follow-up, not this PR: the REPL inline reviewer falls back to the query picker when the direct scheduler says nothing is due (`repl.py` `if card_id is None:`), while `review:next` and `ReviewApp` do not — pre-existing inconsistency, and it may show a card past the daily limit.

PR #69, opened 2026-09-18
feat(backends): one card/note/deck/notetype shape on both backends; fix next-card unit mixing

### LINKED ISSUE #30: Backend Protocol parity: normalize return shapes, lift _store methods, align search semantics

ubermenchh:
Closes #30 — item 1, the last open item.

## Problem

An agent parsing `--format json` had to know which backend it was talking to:

| | direct | AnkiConnect (before) |
|---|---|---|
| notetype on a card | `notetype_name` | `modelName` |
| note id on a note | `id` | `noteId` |
| `fields` | `["front", "back"]` | `{"Front": {"value": …, "order": 0}, …}` |
| `due_info` | decoded (`kind`, `epoch_secs`/`day_index`/…) | absent |
| notetype `id` | present | absent (`modelNames`, not `modelNamesAndIds`) |
| template `ord` | present | absent |

Consumers compensated with dual-shape branches (`render.resolve_notetype_name`, `browse_app._extract_field_values`, the 4-key note-id probe) — and any agent had to do the same.

## Change

- **`models/entities.py`** — `Card`, `Note`, `Deck`, `Notetype`, `DueInfo`: the direct backend's key set, `extra="allow"` (a backend may add keys, never omit one). The Protocol still returns `dict[str, JSONValue]`; the models validate those dicts in tests and document the contract.
- **`backends/normalize.py`** — pure functions mapping AnkiConnect rows onto the canonical keys, **additively**: `notetype_name`, `id`, `fields` as a list, `field_names`, `tags`, `due_info`, `left_info` are added; `modelName`, `noteId`, `question`, `answer`, `css`, `nextReviews` are kept, so existing AnkiConnect scripts keep working. `get_notetypes` uses `modelNamesAndIds`; templates gain `ord`.
- **`core/due.py`** — `decode_due`/`decode_left` moved out of `anki_direct` so both backends share them (re-exported there). AnkiConnect's `due_info` lacks `epoch_secs` for day-index kinds: the collection's rollover timing isn't exposed. New `due_sort_key`.
- **`core/scheduler.py`** — `pick_next_due_card_id` compares by `due_sort_key` instead of raw `due`. The `is:learn is:due` set mixes intraday epochs (~1.7e9) with day-learn day indices (~2e4), so `min(due)` always picked the day-learn card (the #48-review comment on #30). Fixed on both backends.
- Consumers: `browse_app`'s mapping branch and `modelName` fallback deleted; `render.resolve_notetype_name` reads the canonical key first.
- **SKILL.md** — new "Object Shapes" section: card/note/deck/notetype tables, the `due_info` kinds, and a migration note for the AnkiConnect backend.

## Tests

- `tests/integration/test_shape_parity.py`: **both backends through `get_card`/`get_note`, validated against the same model, with equal values wherever AnkiConnect has the data**; direct has `epoch_secs`, AC doesn't (asserted). AC decks/notetypes validate; direct deck/notetype outputs validate in their existing full-DDL tests.
- `tests/backends/test_normalize.py`: each normaliser in isolation, incl. `due_info` for every type/queue combination matching the direct decoding, missing-key tolerance, and that AnkiConnect keys survive.
- `test_scheduler.py`: the unit-mixing counterexample (day index 40 vs epoch 1.7e9) picks by time, and with no epoch for the day-learn card the intraday card still wins.
- `test_docs_sync.py`: every card key in the SKILL table is a `Card` field; `due_info` kinds match `DueInfo`.
- 7 AC backend tests that pinned the raw pass-through updated to the canonical shape; 1 TUI test fixture used the old `modelName` key.

`pytest`: 2061 passed + 89 TUI · `ruff` · `ty` clean.

## Behaviour change (AnkiConnect backend only)

`card`/`cards`/`note` JSON gains keys; nothing is removed. `fields` **changes type** from map to list — that is the one breaking edge, called out in SKILL.md with `field_names` / `note:fields` as the replacements. `decks` items gain `kind: "unknown"`; `notetypes` items gain `id`.

## Noted for #37

This test fixture set's minimal `decks`/`notetypes` DDL can't serve `get_decks`/`get_notetype` (`no such column: common/config`), which is exactly the faithful-DDL problem #37 describes.


ubermenchh:
Review found one real regression, fixed:

- **Unknown id on AnkiConnect returned a zeroed card with exit 0.** AnkiConnect answers `cardsInfo`/`notesInfo` for an unknown id with `{}` rather than an error; the normaliser filled the gaps into a record that validated against `Card`. Now `LookupError` → `ENTITY_NOT_FOUND`/4, same as direct. Parity test drives both backends.
- `due_sort_key` now ranks by **kind** (intraday learning → day-indexed → new) rather than by data availability, so the same collection yields the same `review:next` on both backends; the previous scheme let direct rank a day-learn card ahead of an intraday one while AnkiConnect could not. Kind-first is also Anki's own queue order.
- `get_notetypes` degrades like `get_notetype` on old AnkiConnect; id lookup no longer swallows a dropped connection; normalize docstring names the one non-additive change.

`pytest`: 2064 passed + 89 TUI; ruff / ty clean.

PR #70, opened 2026-09-18
test: shared conftest with Anki's real schema-18 DDL and protobuf-correct row builders

ubermenchh:
Part of #37: the shared fixture, the sync-bookkeeping helpers, and the housekeeping items. The remaining 14 per-file `_make_store` builders migrate in a follow-up PR so this one stays reviewable (the inventory found 20 builders and 11 incompatible `_insert_card` signatures; doing them all here would be a ~2000-line test-only diff).

## What

- **`tests/anki_schema.py`** — the verbatim `sqlite_master` dump of a collection written by Anki Desktop: every `NOT NULL`, `COLLATE unicase`, `WITHOUT ROWID`, and index. Not hand-written.
- **`tests/conftest.py`** — `Collection` with `insert_deck / insert_filtered_deck / insert_deck_config / insert_notetype / insert_note / insert_card / insert_revlog / set_config`, each defaulting every column to a value the store decodes correctly, plus `row / rows / ids / col / graves` readers and `store(writable=)`. Fixtures `collection` (seeded: Default deck 1, Basic notetype 10 with a rendering template, deck_config 1) and `bare_collection`.
- **`assert_synced(table, id)`** — `usn == -1`, row timestamp moved, `col.mod` moved (#47); **`assert_untouched()`** for refusals; **`assert_schema_changed()`** for `scm`. A test forges each defect and checks the helper catches it.

### Why the protobuf defaults matter

| blob | `b""` decodes to | consequence | fixture default |
|---|---|---|---|
| `decks.kind` | kind `"unknown"` | `delete_deck` refuses, `get_deck_config` rejects, move-into-filtered checks fail | normal deck, `config_id=1` |
| `deck_config.config` (row present) | `learn_steps=[]` = "no steps" (#57) | silently changes scheduling in every test | `[1, 10]` / `[10]` |
| `notetypes.config` | normal, sort field 0 | fine, but implicit | explicit |

## Migrated in this PR

`due`, `find_queries`, `card_note_revlog_reads` (adapter signatures kept), `shape_parity`, `answer_card`, `snapshot_restore`. Both cross-file `_make_store` imports are gone.

Things the real schema surfaced immediately:
- `answer_card`'s own `_insert_deck` wrote only `(id, name, kind)` — the exact class of gap the issue describes.
- `answer_card` asserted `"fsrs_params": "unknown"`, which is the store's *missing-table* label; a real collection always has `deck_config`, so with no row the honest value is `"default"`. That assertion was pinning a fixture artefact.
- `shape_parity` gets its direct-side deck/notetype checks back — #69 had to skip them because the old fixture lacked `decks.common` / `notetypes.config`.

Five answer/undo paths now assert `col.mod` moved; the preview-deck refusal asserts it didn't. (The inventory counted ~70 positive-path mutator tests with no `col.mod` check; the rest come with the file migrations.)

## Housekeeping (from the issue)

- `unit` / `slow` markers deleted (never applied); `integration` applied to `tests/integration/` via a collection hook — `pytest -m "not integration"` is now the fast set (1833 tests, 2.5s).
- `test_color_constants_are_hex_rgb` removed; dead first assignment in `test_invoke_non_object_response_raises_protocol` removed.
- SIM118 had already been cleaned up (0 hits).

## Guard

`tests/integration/test_fixture_schema.py` pins every table's exact column list and nullability, key indexes, that the `col` row passes the store's schema guard, and that the seeded blobs decode as normal / `[1,10]` / normal-with-template. If someone "simplifies" the DDL, this fails first.

`pytest`: 2068 passed + 89 TUI · `ruff` · `ty` clean.

## Follow-up

Migrate the remaining builders (`suspend_and_tags`, `deck_write_paths`, `read_apis`, `empty_steps`, `learn_queue_units`, `rollover_timing`, `preview_ratings`, `delete_and_tag_reads`, `write_guard`, `deck_config_and_add_notes`, `card_mutators`, `note_mutators`, `notetype_mutators`, `fsrs_fidelity`) and the CLI helper quartet (8 identical copies) to `tests/cli/conftest.py`, adding `assert_synced` to the ~65 remaining positive-path mutator tests as each file moves.


ubermenchh:
Review: neither risk materialised — no migrated assertion got weaker (the `"unknown"` → `"default"` change verified as a fixture artefact: `"unknown"` is the missing-*table* label), and the fixture defaults check out against how the store decodes each blob.

The faithful DDL earned its keep before merge: Anki declares `sfld integer`, so `get_note()["sfld"]` returned an **int** for a numeric sort field — the API type flipped with the note's content, and no TEXT-column fixture could ever see it. Fixed in the store (`str()`), pinned.

Also applied: `insert_note` rejects bare-string tags Anki never writes; `assert_synced(baseline=)` and a documented precondition on `assert_untouched`; shared `seed_review_card`; the schema self-test now covers the exact table set and every index with its UNIQUE-ness; `item.path`.

`pytest`: 2070 passed + 89 TUI; ruff / ty clean.

PR #71, opened 2026-09-18
test(#37): finish migrating every per-file builder onto the shared Collection fixture

### LINKED ISSUE #37: tests: add conftest.py with faithful Anki DDL; remove 16 store builders and 11 CLI helper copies

ubermenchh:
Closes #37.

Second half of the #37 migration (first half landed in #70). All 14 remaining `_make_store` builders in `tests/integration/` now build on `tests.conftest.Collection` (verbatim schema-18 DDL from `tests/anki_schema.py`). Each file keeps its own helper names/signatures as thin adapters, so test bodies are untouched.

## Files migrated
- suspend_and_tags, delete_and_tag_reads, learn_queue_units, rollover_timing, preview_ratings
- deck_write_paths, deck_config_and_add_notes, card_mutators, empty_steps, fsrs_fidelity
- read_apis, write_guard, note_mutators, notetype_mutators

`rollover_timing`'s `with_config_table=False` now `DROP TABLE config`s after the real schema instead of hand-rolling a partial one. `note_mutators` deliberately seeds notetype 10 with no templates (matching what those tests were written against, the `ensure_not_empty` fallback); `_install_templates` is an adapter over `insert_template`.

`CREATE TABLE` remains only in `anki_schema.py`, the schema-11 `schema_guard`, `parse_decode`'s dummy table, and the `backends/detect*` probes.

## Fixture additions (`tests/conftest.py`)
- `insert_field`, `insert_template`, `insert_notetype(config=, templates=[])` for tests that place rows individually.

## Coverage added while passing through
- `deck_write_paths`, `card_mutators`: `assert_synced` on positive paths; `assert_untouched` on no-op / refused paths.
- `note_mutators`: add/update/delete assert both row `usn=-1` and `col.mod` movement; delete no-op asserts untouched; new parametrized test drives **every** note-side mutator (`add_note`, `update_note` fields/tags, `add_tags`, `remove_tags`, `rename_tag`, `add_notes`) against a row seeded `usn=0` and checks the re-flag + `col.mod` in one place (#47 territory).

## Bonus dedup (review follow-up from #70)
- `tests/cli/conftest.py`: `base_obj` / `success_payload` / `error_payload` / `patch_session` / `failing_session` once; ten files import them under the old underscore names. `error_payload` reads stderr only (per #27). The three `_base_obj`s carrying bootstrap-specific keys (app_bootstrap, config, general) stay local.

## Verification
- `.venv/bin/python -m pytest -q` → 2077 passed
- `.venv/bin/python -m pytest -q -m tui` → 89 passed
- ruff, ty clean


ubermenchh:
Review pass (test-quality reviewer) — one IMPORTANT, reproduced and fixed in 4a7a5e0-ish (latest commit):

- **unbury_cards sync marks were unverified.** Mutant: drop `usn = -1` from both unbury UPDATEs in `anki_direct.py` → card_mutators stayed green, because the only col.mod check after unbury was already satisfied by the preceding bury. Fix: re-arm `col.mod` to baseline before unbury and `assert_synced` on cards bury never touched (3/4; 101/102 in the deck-scope test, 103 pinned untouched). Mutant now fails both tests.
- `note_mutators` `_insert_note`/`_insert_card` seeded `usn=-1`, so `assert_synced`'s usn clause couldn't fail there. Rows now start synced.
- `COL_BASE_MOD_MS`/`COL_BASE_SCM_MS` were duplicated in `tests/integration/conftest.py`; now imported from `tests.conftest` so `assert_col_untouched` can't drift from the seeded baseline.

Reviewer confirmed: no migration weakened an assertion (verbatim DDL is stricter — NOT NULL, `sfld integer` affinity); each branch of the parametrized note-mutator test hits a distinct UPDATE and fails alone when its `usn=-1`/`mod` is removed; `error_payload` reading stderr only can only fail loudly, never pass wrongly (#27 pins the envelope's stream).

Suite: 2077 + 89 tui passed, ruff/ty clean.

ubermenchh:
(Correction: the fix commit is c3214ac.)

PR #72, opened 2026-09-18
refactor(#31): split anki_direct.py into db/ modules; break the db -> backends cycle

### LINKED ISSUE #31: Split anki_direct.py (2.9k lines) into db/ modules; fix db -> backends import cycle
## Problem

`anki_cli/db/anki_direct.py` is 2,929 lines, one class (`AnkiDirectReadStore`), ~90 methods, and it performs every write in the project despite the "ReadStore" name. It also imports upward: `:2224` pulls private helpers `_anki_process_running`, `_sqlite_write_locked` from `backends/detect.py`, and `:17` imports `core.search`, which is really a SQL compiler with exactly one caller (this file). So the actual dependency graph is `cli → backends → db → core` **plus** `db → backends`, a cycle.

Duplicated blocks inside the file:
- `_allocate_epoch_ms_id` (`:2232-2248`) is byte-identical to `_allocate_row_id`.
- ~40 lines of FSRS seeding duplicated verbatim (`:933-1000` vs `:1896-1964`).
- The `queue = CASE …` SQL appears three times (`:1088-1099`, `:1113-1124`, `:2393-2408`).
- `csum` computed twice from the same input (`:1553`, `:1588`).
- Inline `import sys` / `import re` (`:1562`, `:2336`).

`backends/direct.py` is a 200-line pure delegation class over the store; `direct.py:19-23` re-resolves and re-checks a path the store's `__init__` immediately checks again.

## Suggested layout

The file's own section comments already map onto this split, so it's mechanical:

```
anki_cli/db/
  lock.py         _anki_process_running, _sqlite_write_locked  (imported by both detect.py and the store)
  connection.py   _connect, _connect_write, _ensure_write_safe, _allocate_id, _day_index, col.mod/scm bump
  codec.py        proto decode helpers, tags/fields split+join, csum/sfld (HTML-stripping), card data JSON
  search_sql.py   (moved from core/search.py)
  decks.py        get/create/rename/delete/config  (+ filtered-deck awareness)
  notetypes.py    create/field/template/css        (+ flds rewrite, scm bump)
  notes.py        add/update/delete/tags           (+ reqs-based card generation)
  cards.py        read, suspend/bury/move/flag/reset/reschedule
  scheduling.py   FSRS build/seed/map, answer_card, preview_ratings, snapshot/restore
  store.py        class AnkiDirectStore(DecksMixin, NotetypesMixin, ...)
```

- Rename `AnkiDirectReadStore` → `AnkiDirectStore`.
- `core/` keeps only genuinely backend-independent code (`template.py`, `scheduler.py`; `undo.py` is direct-only and could move too).
- Either have `AnkiDirectStore` implement `AnkiBackend` directly and delete `backends/direct.py`, or shrink it to a 20-line facade.

Do this **after** the data-safety fixes land (they're small diffs against the current file), and before adding any more features to the store.

ubermenchh:
Closes #31.

## What
Three commits, reviewable independently:

1. **Break the cycle** (`29807c3`). `db/lock.py` takes `anki_process_running` (+ per-platform probes) and `sqlite_write_locked` out of `backends/detect.py`; `core/search.py` becomes `db/search_sql.py` (its only caller was the store). `db/` now imports nothing from `backends`/`cli`/`tui`. The store calls the probes as attributes of the `lock` module on purpose, so the write-guard tests stub them in one stable place.

2. **Split the store** (`c0e19dc`). Pure move of the 4.1k-line file into one module per area. Each is a mixin that *extends the mixins it calls into*, so every `self.<other area>` call is type-checked (`ty` clean) rather than assumed:

   ```
   CodecMixin                 codec.py       protobuf/field/tag/checksum/card-data, no I/O
     ConnectionMixin          connection.py  __init__, open/write txn, ids, timing, col.mod/scm
       DeckLookupMixin        deck_lookup.py name -> id, subtree filter, deck options
         NotetypesMixin       notetypes.py
         CardsMixin           cards.py       + RESTORED_DUE_* / queue_from_type_sql
           DecksMixin         decks.py       (above cards: get_deck reports due counts)
           SchedulingMixin    scheduling.py  FSRS answer/preview/snapshot
         NotesMixin           notes.py
   AnkiDirectStore(DecksMixin, NotesMixin, SchedulingMixin)   store.py
   errors.py                  the typed refusals
   ```
   Method bodies are byte-for-byte the originals (AST-verified against `main`; the only exceptions are the two `_ensure_write_safe`/`_field_checksum` reference changes and the commit-3 extraction below) (`ruff format` was deliberately **not** run: the repo isn't formatted and CI only runs `ruff check`).

   `AnkiDirectReadStore` → `AnkiDirectStore`. Public methods now take the `AnkiBackend` protocol's signatures (positional where the protocol is positional, same defaults; `find_note_ids`/`find_card_ids` → `find_notes`/`find_cards`), so `backends/direct.py` shrinks from a 234-line delegation class to `DirectBackend(AnkiDirectStore, AnkiBackend)` carrying only `name` / `supports_scheduler_introspection` / `collection_path`. The duplicate path-exists check goes with it; the store now raises the message the backend used to ("Direct collection not found", previously "Direct DB not found" one layer down).

3. **Dedupe** (`73db372`). `answer_card` and `preview_ratings` shared ~60 identical lines (timing, preview-deck refusal, home-deck scheduler, card→py-fsrs mapping, revlog/legacy seeding); now one `_prepare_fsrs_review` returning a NamedTuple, so the preview cannot drift from the write. `_allocate_epoch_ms_id` (byte-identical to `_allocate_row_id`) removed.

   Of the issue's other dedupe items, the `queue = CASE` triplication, double `csum`, and inline imports had already been fixed in earlier PRs.

## Not done
- `core/undo.py` stays: it imports nothing from `db` and is a CLI-level JSON journal.

## Tests
- `direct_mod.*` references now point at the owner of the name: `time`/`datetime` patched on the stdlib module or `anki_cli.db.scheduling`; fsrs names imported from `fsrs`; lock probes stubbed on `anki_cli.db.lock`.
- Probe tests move to `tests/db/test_lock.py`.
- The 45-row delegation table in `test_direct_backend.py` is replaced by a signature-compatibility test over every protocol method (positional-ness, defaults, no extra required params).
- 2074 + 89 tui passed; ruff, ty clean. (Count vs main: `test_direct_backend.py` went from 49 collected (2 + a 47-row delegation table) to 47 (3 + one signature check per each of the 44 protocol methods); the 16 probe tests moved 1:1 to `tests/db/test_lock.py`.)


umangk-2410:
Review pass (architecture + correctness reviewers; both ran without a shell, so I executed their mechanical checks myself):

**Body identity vs `main`** — AST-compared every method body and top-level definition in `db/*.py` against `origin/main:anki_cli/db/anki_direct.py`. The only bodies that differ are the intended ones: `_ensure_write_safe` (lock module refs, commit 1), `_field_checksum` (`AnkiDirectReadStore.` → `CodecMixin.`), `_allocate_row_id` (docstring), `answer_card`/`preview_ratings` (extraction). Every top-level constant/regex/function is identical. Signature changes are exactly the protocol alignment listed in the PR body.

**`_prepare_fsrs_review`** — normalised diff against *both* original blocks: identical modulo the `home_did` alias and trailing commas. `review_dt` is still taken where each caller took it (preview: before the SELECT; answer: after), and `_timing` still runs before the preview-deck refusal in both, as before.

**Findings acted on** (commit above): `collection_path` is now a plain attribute matching the protocol declaration; `store.py` layering docstring fixed (NotetypesMixin extends ConnectionMixin); `strip_html_preserving_media_filenames` de-underscored for the cross-module import; blank-line tidy.

**Found but deliberately not fixed here** — `rename_deck`/`delete_deck`/`unbury_cards` build the deck-subtree query with `LIKE` and no `ESCAPE`/`NOCASE`, unlike `_deck_filter`. Real (`delete_deck("A_B")` also deletes `AXB::*`), but pre-existing and behaviour-changing → #73.

**Declined**: turning the self-less helpers into module functions and moving revlog decoders to `codec.py` (the reviewer's own verdict was "do not restructure again"; can be a follow-up), inlining `_build_scheduler_ex`.

Suite: 2074 + 89 tui, ruff/ty clean.

ubermenchh:
Second review pass (pr-reviewer + test-quality-reviewer), one fix commit:

- **IMPORTANT, fixed** — the new signature test resolved methods on `DirectBackend`, so a store method that went missing fell through the MRO to the protocol's own stub and compared equal to itself. It also never checked positional *order*. Now resolves on `AnkiDirectStore`, asserts the function isn't the protocol's, pins order, guards against an empty parametrize, and makes one real call through a `DirectBackend`. Verified by mutation: renaming `find_notes` and swapping `rename_deck(old_name, new_name)` each fail exactly one case.
- **Body corrections** — the test-count arithmetic was wrong (44 protocol methods, not 46/47); "bodies identical" now names its exceptions; the `"Direct DB not found"` → `"Direct collection not found"` message change is disclosed.
- **Commit hygiene** — c0e19dc does carry the signature alignment alongside the move; its message enumerates it, and the squash merge will carry the PR body. Not re-splitting.
- Coverage delta confirmed by the reviewer: 16/16 probe tests present in `tests/db/test_lock.py`; `detect_backend`'s own probe tests still stub `detect_mod` aliases; every rewired monkeypatch target (`time`, `scheduling_mod.datetime`, `connection_mod.local_minutes_west_for_stamp`, `lock_mod.*`) verified live against the production call sites.

2075 + 89 tui, ruff/ty clean.

PR #74, opened 2026-09-19
fix(#73): deck-subtree matching is literal and unicase everywhere

### LINKED ISSUE #73: Deck-subtree matching: rename_deck/delete_deck/unbury_cards use LIKE without ESCAPE or NOCASE
Found during the #31 review (PR #72). Pre-existing; not fixed there because the PR is a pure move.

`_deck_filter` (`anki_cli/db/deck_lookup.py`) is the one correct implementation of "a deck and its children": `name = ? COLLATE NOCASE OR name LIKE ? ESCAPE '\'` with `escape_like()`. Three other places hand-roll the same query without either safeguard:

- `anki_cli/db/decks.py` `rename_deck` (source scope **and** the target-conflict probe)
- `anki_cli/db/decks.py` `delete_deck`
- `anki_cli/db/cards.py` `unbury_cards(deck=...)`

Consequences:
- `_` and `%` in a deck name are LIKE wildcards. `delete_deck("A_B")` also deletes `AXB::*` (and their cards); `rename_deck` moves the wrong subtree; `unbury_cards(deck="A_B")` unburies `AXB::*`.
- No `NOCASE`: Anki treats deck names as case-insensitively unique, so `delete_deck("japanese")` misses `Japanese` while `--deck japanese` (via `_deck_filter`) finds it.

Fix: add `_deck_subtree_ids(conn, name) -> list[int]` (or reuse `_deck_filter`) in `deck_lookup.py`, have all four callers use it, and pin with tests: a deck literally named `A_B` next to `AXB::child`, and a case-only mismatch. `rename_deck`'s conflict probe needs the same treatment for the *target* name.

ubermenchh:
Closes #73.

## Bug
`rename_deck`, `delete_deck`, `unbury_cards(deck=)` and `_deck_filter` each built "the deck and its children" by hand. Three used `LIKE '<name>::%'` with no `ESCAPE`, so `_`/`%` in a deck name were wildcards:
- `delete_deck("A_B")` also deleted `AXB::child` **and its cards**
- `rename_deck("A_B", ...)` moved `AXB::child` too
- the rename target-conflict probe refused `A_C` if `A_B` existed (LIKE-match), while a legitimate case-only rename could collide with itself

`LIKE` is also ASCII-case-insensitive only, so `unbury_cards(deck="école")` missed `École::Verbs`. And `_deck_filter`'s explicit `COLLATE NOCASE` was a *downgrade*: `decks.name` is `COLLATE unicase` (the store registers Anki's collation), so plain `=` already folds non-ASCII; `--deck école` returned nothing while `École` existed. (Correction to the issue text: the `=` half was never the problem; the explicit `NOCASE` was.)

## Fix
`DeckLookupMixin._deck_subtree(conn, name) -> DeckSubtree | None` is the one implementation. Parent matched with plain `=` (column unicase, exactly Anki's uniqueness rule); descendants matched with `substr(name, 1, n) = '<stored parent>::'`, so the prefix is literal and no escaping is needed. Returns the stored parent name; `rename_deck` slices suffixes from that rather than from the caller's casing, and treats a target that resolves to the decks being renamed as not-a-conflict (case-only rename works).

All four callers route through it. No `LIKE` on deck names remains in `db/`.

## Tests
New `tests/integration/test_anki_direct_deck_subtree.py` (23 cases): the helper against `A_B`/`AXB`, `100%`/`100x`, `École`, `Base`/`Basement`, canonical name + parents-first order, `None` for missing; then `delete_deck`, `rename_deck` (scope, target probe, case-only rename), `unbury_cards`, and `_deck_filter` via `get_due_counts`. All 23 failed on `main` (except the two `None`-for-missing cases).

2098 + 89 tui passed; ruff/ty clean.


ubermenchh:
Review pass (correctness reviewer) — two IMPORTANT and one pre-existing, all acted on:

- **Descendant match was BINARY** (`substr(...) = prefix` drops the column collation). A child written as `lang::De` under `Lang` (which `write_deck` allows) fell out of the subtree — a regression from the old ASCII-folding `LIKE`. Now uses rslib `child_decks`' range `name > 'P::' AND name < 'P:;'`, which runs under the column's unicase collation and still has no wildcards. Added `Lang`/`lang::De` cases.
- **Two tests passed on the old code** (target `A_C` was never LIKE-matched; `École`→`ÉCOLE` matched byte-identically). Now `AX_` (LIKE-matched `AXB::child`) and `école`→`ÉCOLE` (ASCII `LIKE` missed the child). I re-ran all eight caller tests against `main`'s three files: all fail with the bug's symptoms (`3 == 2` over-deletion, a `UNIQUE` collision from the rename over-match, `unburied 2 != 1`, `_deck_filter` `0 == 1`).
- **Pre-existing, filed as #75**: Anki stores `\x1f` between deck-name components; the direct store assumes `::` everywhere (this PR keeps that assumption, so it's unchanged here). Verified on a copy of my collection that the `::` rows present were written by this CLI.

2100 + 89 tui, ruff/ty clean.

PR #76, opened 2026-09-19
fix(#32, part 1): lazy backend detection, single version probe, lock before pgrep

ubermenchh:
Part 1 of #32 (detection). Part 2 (AnkiConnect client hygiene, fallback removal, `multi` batching) follows.

## Before
Every subcommand paid for detection in the Click group callback — 0.35 s HTTP probe, a `pgrep` subprocess, a SQLite lock probe — before the command had decided whether it needed a backend; then `factory.py` posted `version` to AnkiConnect a **second** time with a fresh client. Both direct paths ran `pgrep` before the cheaper, authoritative lock probe.

(Already fixed on `main` before this PR, from the issue's list: `standalone` fallback → `DetectionError`, missing `--col` → error, `anki_profile` honoured, backendless commands skip detection.)

## After
- **Lazy, once.** `app.py` records the requested backend with `backend_reason = "pending"`; `create_backend_from_context` detects on first use and writes `backend` / `collection_path` / `backend_reason` / `ankiconnect_version` back into `ctx.obj`, so the REPL's per-action sessions never re-probe. Failure is cached as `backend = "none"` + reason. Bare `anki` (REPL) keeps eager detection: it shows the backend in its header and must warn up front.
- **Exit codes preserved.** `BackendFactoryError` carries the detection's exit code (3 nothing found / 7 unavailable); `backend_unavailable` and `classify()` honour it.
- **One `version` round trip.** `_ankiconnect_version` returns the probed version; the factory passes `verify_version=False` when it's an int ≥ 6, otherwise lets `check_version` explain (too old / malformed).
- **Lock before pgrep** in both direct paths.
- `OutputFormatter` reads `meta.backend`/`meta.collection` from the ctx object at emit time (the formatter now exists before resolution).

## Behaviour changes (user-visible, intended)
- A detection failure's envelope now has `meta.command = "<the command>"` instead of `"bootstrap"`, and `details = {"backend": "none", "requested_backend": "<auto|direct|ankiconnect>"}` instead of `{"forced_backend": ...}`. Exit codes unchanged (verified end-to-end: forced-unreachable → 7, missing `--col` → 3).
- Commands that never open a session (a future `--help`-like path, or a command that fails input validation first) no longer probe anything.

## Tests
- `test_app_bootstrap.py`: detection-error exit code survives via the factory; a subcommand that never opens a session never detects; first session detects once and caches; failed detection cached; probed version skips `verify_version` (and doesn't for < 6 / hand-built obj); `anki_profile` still reaches the lazy call site.
- `test_detect.py`: lock probe refuses without ever spawning the process scan (both paths); result carries the probed version.
- `test_detect_helpers.py`: `_ankiconnect_version` for int / non-int / error payloads.
- `test_formatter.py`: `meta` reflects post-construction resolution; constructor-only formatter unchanged.

2115 + 89 tui passed; ruff/ty clean.


ubermenchh:
Review pass (correctness reviewer) — two IMPORTANT, both real, fixed in the commit above:

- **`anki shell` took the lazy path.** On a host with no Anki it opened the REPL silently (bare `anki` warned). Both REPL launchers (`None`, `"shell"`) now detect eagerly; test parametrised over both.
- **`meta.backend = "auto"` on pre-session envelopes.** A command failing input validation never opens a session; the formatter read the stored *preference*. While pending it now reports `"none"` (the documented no-backend value). Reproduced end-to-end before/after with `anki --format json --col /nonexistent deck --deck ""`.
- Nits: cached failure keeps its exit code (3 stayed 3, not 7, on the second session); `forced_backend` no longer falls back to `obj["backend"]`.
- Not changed: probe reordering alters only the *message* when `--col` is unopenable *and* Anki is running ("Cannot probe the lock state…" instead of "Anki is running…") — more accurate, exit code identical, no test depended on it.

2119 + 89 tui, ruff/ty clean.

ubermenchh:
Correction to the comment above: the previous push had 19 red tests in `test_app_error_envelope.py` (they pinned `meta.backend == "direct"` for commands that raise before opening a session — exactly the case the fix changes to `"none"`). I chained the commit behind the suite with `;` instead of gating on it; my mistake. Fixed in the follow-up commit: assertions updated with the reasoning, format assertions unchanged. Now 2117 + 89 tui green locally.

PR #77, opened 2026-09-19
fix(#32, part 2): AnkiConnect client hygiene, documented request shapes, multi batching

### LINKED ISSUE #32: Lazy backend detection + AnkiConnect HTTP hygiene (Connection: close, timeouts, N+1, param fallbacks)

ubermenchh:
Part 2 of #32; closes #32 together with #76.

Every request shape below was checked against the current upstream v6 API docs (git.sr.ht/~foosoft/anki-connect).

## Transport
- **Connection reuse.** The pooled `httpx.Client` was created for reuse, then every request sent `Connection: close`. Gone. A `RemoteProtocolError` — AnkiConnect closes idle keep-alive sockets, and a request written onto one that just died fails *before* Anki sees it — is retried once on a fresh connection — but only for reads and idempotent writes (allow-list; `multi` only if every sub-action qualifies), because the same error also means "Anki executed it, then died before answering". Non-idempotent actions fail with "Not retrying 'addNote': check in Anki whether it applied."
- **Split timeouts.** `connect=2 s`, `read=write=30 s` (`read_timeout_seconds` is a new constructor arg). A `findCards` over a large collection no longer surfaces as "Timeout contacting AnkiConnect".

## Request shapes
- `updateModelTemplates` / `updateModelStyling`: the documented `model={name, …}` shape, once. The old first attempt failed every time and the `except AnkiConnectAPIError` swallowed real errors (e.g. "model was not found").
- `guiAnswerCard`: `ease`. The `answerEase` retry was invented.
- **`renameDeck` does not exist.** The three spellings that always failed are gone; the create/changeDeck/deleteDecks emulation *is* the implementation — and review found it never worked either: `deleteDecks` rejects `cardsToo=false` (Anki ≥ 2.1.28), so every rename ended half-applied, and `deck:"Old"` includes subdecks, so children's cards were swept into the parent's new deck. Now: move each deck's *own* cards, prove the old subtree is empty, delete with `cardsToo=true`. All name comparisons casefold (Anki resolves names case-insensitively); an occupied target, a case-only rename, and a target inside the source subtree are refused up front with the reason.
- **`bury` / `unbury` / `unburyCards` do not exist.** `bury_cards` / `unbury_cards` raise `BackendUnsupportedError` (→ `BACKEND_UNAVAILABLE`, exit 7, "Use --backend direct.") without a request. They had never worked over AnkiConnect (exit code moves from 1 to 7). README/SKILL examples annotated.
- `modelNamesAndIds` / `modelStyling` pre-v6 fallbacks removed; `check_version` already pins ≥ 6.

## Round trips (new `_multi` helper)
Sub-actions are labelled with the API version so the server envelopes each answer; a failing sub-action raises `AnkiConnectAPIError` naming *that* action. Malformed batch responses (short, not a list, unenveloped items) are protocol errors.

| method | before | after |
|---|---|---|
| `get_notetypes` | 1 + 2·models | 2 |
| `get_notetype` | 4 | 1 |
| `get_tag_counts` | 1 + tags | 2 |
| `get_due_counts` | 3 | 1 |
| `set_card_flag` | cards | 1 (every failed card named; the batch ran regardless) |

## Tests
- Transport: no `Connection: close` header; timeout split; retry exactly once on `RemoteProtocolError`, twice → unavailable; `ReadTimeout` not retried.
- `_multi`: version labelling + ordered unwrap, empty batch makes no request, sub-action error names the action, three malformed shapes rejected.
- Each rewired method pins its exact `multi` payload; `rename_deck` pins the emulation sequence, occupied-target refusal (`Taken`, `Taken::Sub`, ancestor), own-subtree refusal, and that no `renameDeck` is ever sent.
- `test_shape_parity`'s fake `_invoke` dispatches `multi` like the server.

2137 + 89 tui passed; ruff/ty clean.

Deferred to a possible part 3: `answerCards` (non-GUI; would lift the "must be the card Anki is showing" restriction) and `getReviewsOfCards` for `get_revlog`.


ubermenchh:
Review pass (correctness reviewer, with the v6 README as oracle) — one BLOCKING, two IMPORTANT, all fixed in the commit above:

- **BLOCKING — `deleteDecks(cardsToo=False)` is rejected** ("The argument `cardsToo` *must* be specified and set to `true`", Anki ≥ 2.1.28). Every `rename_deck` over AnkiConnect ended half-applied: new decks created, cards moved, then an API error with the empty old deck left behind. Pre-existing, but this PR made the emulation the only path. While fixing it I found the emulation also used `deck:"Old"` per deck, which in Anki *includes subdecks* — so the parent's move swept the children's cards into `New` instead of `New::Child`. Now: move each deck's own cards (`deck:"X" -deck:"X::*"`), prove the whole old subtree is empty with one `findCards`, then delete with `cardsToo=true`; if cards remain, stop before any delete with an actionable message.
- **IMPORTANT — case-sensitive conflict checks.** Decks `Old`, `taken`; `rename_deck("Old","Taken")` passed the check, `createDeck("Taken")` resolved to the existing `taken`, cards silently merged. All comparisons now casefold; case-only renames are refused (`createDeck` returns the same deck, the delete would then remove the cards); suffixes slice from the stored spelling.
- **IMPORTANT — blind retry on `RemoteProtocolError`.** Replaying `addNote` duplicates, `guiAnswerCard` rates the next card. Retry is now gated on an allow-list of reads + idempotent writes (`multi` only if every sub-action qualifies); anything else fails with "Not retrying '<action>': check in Anki whether it applied."
- `set_card_flag` reports every failed card (the batch ran regardless); dead `_timeout_seconds` removed.
- Disclosed: `card:bury`/`card:unbury` on AnkiConnect now exit 7 (`BACKEND_UNAVAILABLE`) instead of 1 — they never worked, this just says so by type.

2148 + 89 tui, ruff/ty clean.

